PR #2, opened 2015-05-05
Updated readme to standard install info

rbillings:


README.md
```diff
@@ -1,2 +1,73 @@
...
+
+__note__
+
+
+
+### Running tests locally
+
+To run tests locally, it's a simple case of calling the command below from this directory
+
+    py.test --driver=firefox
```
tojonmz: I wonder if we should add `--baseurl=https://https://treeherder.allizom.org` here? I get an error baseurl is missing when I run otherwise.

README.md
```diff
@@ -1,2 +1,73 @@
-# treeherder-tests
-Website automation for Treeherder https://github.com/mozilla/treeherder-service
+Treeherder Tests
+============================
+Website automation for Treeherder for [https://treeherder.mozilla.org/][MOZ]
+
+[MOZ]: https://treeherder.mozilla.org/
```
tojonmz: I'm wondering for L3-5, we might have something like:
```
Website automation for Treeherder reporting dashboard [MOZ]

[MOZ] Production: https://treeherder.mozilla.org/
[MOZ] Stage: https://treeherder.allizom.org/
```

README.md
```diff
@@ -1,2 +1,73 @@
...
+to install the required Python libraries.
+
+__note__
+
+The instructions below will install the required Python libraries into your
+global Python installation. If you work on multiple Python projects that might
+end up needing different versions of the same libraries, you might want to
+follow `sudo easy_install pip` with `sudo pip install virtualenv`, and then
+create and activate a [virtualenv](http://www.virtualenv.org) (e.g. `virtualenv
+mcom-tests-env; source mcom-tests-env/bin/activate`) to create a clean
```
tojonmz: Fwiw, wondering if we might consider changing this to something like:

"...and then create and activate a virtualenv in the project root by typing:
```
user@machine: virtualenv .venv
user@machine: source .venv/bin/activate
```

The `.` dot venv above, is so *py.test* doesn't find other tests other than in the ./tests folder. I didn't realize this based on the example 'mcom-tests-env' (so I used 'venv'), and py.test was finding hundreds of other tests in venv, excluding them and also complaining about missing os and sys modules when I ran the tests in the project root rather than from the tests directory:
```
>               raise ImportError("No module named '%s'" % (name,))
E               ImportError: No module named 'pip._vendor.pkg_resources.tests.__builtin__'
```

Or tweaking the above to whatever you think makes sense. :)

rbillings: I edited it to call out those lines, as that makes sense.

README.md
rbillings: I added line 3 to call out being able to use baseurl, and that stage is the default. I also added the mozwebqa.cfg file which will actively do that without needing to call out the baseurl.

Issue #1, opened 2015-05-05
Verify unclassified failure


rbillings:
Find unclassified failure with a parsed log status, verify the log opens in a new tab with data.

rbillings:
This is complete.

PR #3, opened 2015-05-05
Created requirements file and initial tests to find job failure and open the log

rbillings:
Standard Web QA requirements

### DIFF
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -13,32 +13,32 @@ from pages.base import Base
 
 class ResultsetPage(Base):
 
-    _job_log_locator = (By.CSS_SELECTOR, '.logviewer-icon')
-    _job_resultpane_status = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span:nth-child(2)')
-    _log_viewer_locator = (By.CSS_SELECTOR, 'logviewer-icon')
+    _job_details_locator = (By.ID, 'result-status-pane')
+    _job_details_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span:nth-child(2)')
+    _logviewer_locator = (By.ID, 'logviewer-btn')
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
-    _result_status_locator = (By.ID, 'result-status-pane')
+
 
     @property
-    def job_resultpane_status(self):
-        self.selenium.find_element(*self._result_status_locator)
-        return self.selenium.find_element(*self._job_resultpane_status).text
+    def job_details_status(self):
+        self.selenium.find_element(*self._job_details_locator)
+        return self.selenium.find_element(*self._job_details_status_locator).text
 
     def go_to_page(self):
         self.open('')
 
-    def open_next_failed_job(self):
+    def open_next_unclassified_failure(self):
         WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._resultset_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("n")
 
-    def open_job_log(self):
-        self.selenium.find_element(*self._result_status_locator)
-        self.selenium.find_element(*self._job_log_locator).click()
+    def open_logviewer(self):
+        self.selenium.find_element(*self._job_details_locator)
+        self.selenium.find_element(*self._logviewer_locator).click()
 
-    def select_log_viewer(self):
-        self.selenium.find_element(*self._log_viewer_locator).click()
+    def select_logviewer(self):
+        self.selenium.find_element(*self._logviewer_locator).click()
 
-class JobLogPage(Base):
+class LogviewerPage(Base):
 
     _page_title = 'Log for'
     _job_header_locator = (By.CSS_SELECTOR, '.job-header')
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -13,16 +13,14 @@ from pages.base import Base
 
 class ResultsetPage(Base):
 
-    _job_details_locator = (By.ID, 'result-status-pane')
-    _job_details_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span:nth-child(2)')
+    _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
-
+    _result_status_locator = (By.ID, 'bottom-left-panel')
 
     @property
-    def job_details_status(self):
-        self.selenium.find_element(*self._job_details_locator)
-        return self.selenium.find_element(*self._job_details_status_locator).text
+    def job_result_status(self):
+        return self.selenium.find_element(*self._job_result_status_locator).text
 
     def go_to_page(self):
         self.open('')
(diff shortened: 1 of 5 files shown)


rbillings:
This addresses issue #1

pages/resultset.py
```diff
@@ -0,0 +1,59 @@
...
+from pages.base import Base
+
+
+class ResultsetPage(Base):
+
+    _job_log_locator = (By.CSS_SELECTOR, '.logviewer-icon')
+    _job_resultpane_status = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span:nth-child(2)')
+    _log_viewer_locator = (By.CSS_SELECTOR, 'logviewer-icon')
+    _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
+    _result_status_locator = (By.ID, 'result-status-pane')
```
tojonmz: Good stuff! I had to think about this a bit also. Assuming `locator` is a convention, here's a tweak if you think makes sense. We use "job details" to generally describe the entire panel which opens when a job is selected. Note I have slightly re-ordered them:

```
_job_details_logviewer_icon_locator = (By.CSS_SELECTOR, '.logviewer-icon')
_job_details_logviewer_btn_locator = (By.CSS_SELECTOR, 'logviewer-icon')
_job_details_result_status_pane = (By.ID, 'result-status-pane')
_job_details_result_status = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span:nth-child(2)')
_resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
```

So we could generally use these terms I think, for the main UI:
\_navbar (main navbar)
\_navbar_quickfilter
\_resultset
\_resultset_revisions (left side)
\_resultset_jobs (right side)
\_resultset_bar (the per resultset bar containing Date/Author/SHA and action menu dropdown at the right side)
\_job_details
\_job_details_navbar
\_job_details_actionmenu (the left side buttons with log icon, rawlog icon, pin icon, retrigger icon, etc)
\_job_details_tab_\[name\] (the tabs "job details", "failure summary", etc)
\_pinboard
\_logviewer
\_logviewer_header
\_logviewer_failuresteps
\_logviewer_body
\_help

If the above is too granular, feel free to turf what isn't needed :)

pages/resultset.py
```diff
@@ -0,0 +1,59 @@
...
+        self.selenium.find_element(*self._resultset_locator).send_keys("n")
+
+    def open_job_log(self):
+        self.selenium.find_element(*self._result_status_locator)
+        self.selenium.find_element(*self._job_log_locator).click()
+
+    def select_log_viewer(self):
+        self.selenium.find_element(*self._log_viewer_locator).click()
+
+class JobLogPage(Base):
```
tojonmz: Similarly, wondering if we can use `LogviewerPage` if you think it makes sense.

tests/test_unclassified_job.py
```diff
@@ -0,0 +1,38 @@
...
+    @pytest.mark.nondestructive
+    def test_open_failed_job_log(self, mozwebqa):
+        # Open the job log and verify there is content
+        resultset_page = ResultsetPage(mozwebqa)
+        resultset_page.go_to_page()
+        resultset_page.open_next_failed_job()
+        joblog_page = resultset_page.open_job_log()
+        joblog_page = JobLogPage(mozwebqa)
+
+        Assert.true(joblog_page.is_job_status_visible)
```
tojonmz: Just noticed here the newline character (per red icon above) seems to be missing.

tojonmz:
I think best practice git message length is 50/72, per Github's page [here](https://github.com/blog/926-shiny-new-commit-styles) and on [StackOverflow](http://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting). I think we're more lax and generally strive for 72, but maybe we could shorten and call this commit something like:

`Test Case - Select failure and open logviewer (#1)`

I included the Issue number in the message, which does that whizzy Github cross referencing of the PR and the Issue. Or whatever you think is best :)

pages/resultset.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def job_resultpane_status(self):
+        self.selenium.find_element(*self._result_status_locator)
+        return self.selenium.find_element(*self._job_resultpane_status).text
+
+    def go_to_page(self):
+        self.open('')
+
+    def open_next_failed_job(self):
```
tojonmz: Probably best to call this `next_unclassified_failure` in keeping with our treeherder code base. The 'n' key will only find those types. After sheriffs classify them the 'n' key would not longer hit them, though they are also thought of as a failed job. I know.. subtleties abound in treeherder :)

pages/resultset.py
rbillings: Locators are a convention, as is listing them in alphabetical order. Check out my changes and see what you think.
I have updated the files to refer to resultset, job_details, and logviewer. I’ll keep the UI suggestions for later also.

pages/resultset.py
rbillings: Makes sense

pages/resultset.py
rbillings: Done!

rbillings:
@tojonmz - There is no rule about having a correspondence between the id and the names. Ideally it would be identical to reduce confusion. Would a common user be confused by referring to the entire lower panel as "job details" vs "result-status-pane"? I wouldn't have known that the left vs right vs entire panel was an issue just looking at the site- and I couldn't find details in the Help either.

rbillings:
@m8ttyB - r+?

tojonmz:
Our problem in Treeherder is we have a 'job details' (the entire lower panel), a 'job details tab' (which exists within that panel), and a 'job details pane' (the lower left pane). The latter term is more my terminology than other Treeherder users, in an attempt to have a differentiation.

Mostly Treeherder folks have always referred to that lower left pane (id' as `result-status-pane`) as the 'job details' and simultaneously referring to the entire panel as the same thing.

It hasn't cropped up until now as far as reconciliation. I just figured it would help when users are looking at test code to easily grok them. Maybe it's enough they just have to look at the selector to understand what it is.

m8ttyB:
In response to @tojonmz's musings .. it's definitely important to know who the audience is that will be reading, maintaining, and creating testcases. For the typical outsider, like myself, domain specific knowledge about names is difficult. I prefer information that is readily available in the `dom`. 

That said perhaps both worlds can be satisfied by using code comments here.

tests/test_unclassified_job.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @pytest.mark.nondestructive
+    def test_unclassified_failure(self, mozwebqa):
+        # Open resultset page and search for next unclassified failure
+        resultset_page = ResultsetPage(mozwebqa)
+        resultset_page.go_to_page()
+        resultset_page.open_next_unclassified_failure()
+
+        teststatus = resultset_page.job_details_status
+        Assert.equal(teststatus, 'testfailed')
```
tojonmz: I think we might need to assert `'testfailed'` or `'busted'` (red job) here. With unclassified failure navigation (n,p shortcuts), the first job we land on could be either of those types. So if we match either one of them we should pass.

tests/test_unclassified_job.py
```diff
@@ -0,0 +1,38 @@
...
+class TestUnclassifiedJobs:
+
+    @pytest.mark.nondestructive
+    def test_unclassified_failure(self, mozwebqa):
+        # Open resultset page and search for next unclassified failure
+        resultset_page = ResultsetPage(mozwebqa)
+        resultset_page.go_to_page()
+        resultset_page.open_next_unclassified_failure()
+
+        teststatus = resultset_page.job_details_status
```
tojonmz: Running locally I kept getting failures on both tests, but I put a print in here and it appears that this value is empty for some reason. I'll dig around more when I get the chance.

rbillings:
I updated the PR to include testfailed or busted as a status. Please re-run and see if the test passes for you.

tojonmz:
Sweet, I just tried the updated PR locally and it's finding an unclassified failure on stage now and passing both tests (the first failure happened to be an orange aka. 'testfailed' today on stage).

rbillings:
@m8ttyB - final r+? We got the locators sorted out, as well as the additional test status variables.

m8ttyB:
r+ :)

tojonmz:
\o/ thank you @rbillings and @m8ttyB !

Issue #14, opened 2015-07-13
Confirm a failure is available before starting test


tojonmz:
I noticed this issue running our first proof of concept test_unclassified_job on treeherder stage today.

So in https://github.com/mozilla/treeherder-tests/blob/master/pages/resultset.py#L18 we check to see if a resultset has loaded, and if it has loaded proceed with the test.

Unfortunately for unclassified failure navigation, the first unclassified failure may not exist in the first resultset, it may exist in another resultset farther down the page, which may not have loaded yet. And `result-set` is replicated across each resultset eg.

![resultset1vs2](https://cloud.githubusercontent.com/assets/3660661/8656154/e9c9dcd8-2963-11e5-8a17-80a0af64e0b0.jpg)

So we need to either wait or otherwise figure out a way to confirm an unclassified failure is available in the dom before issuing the 'n' to navigate to it.

**nb.** there's a reasonable expectation stage should always have at least one unclassified failure in mozilla-inbound, so we're ok from that standpoint.

tojonmz:
So perhaps one way we could do this is if we can loop every (n)ms with some timeout, over the unclassified failure button (top right), the moment the child span value is >0 we break out of that loop and issue the 'n' key, since we know there is at least one failure in the dom.

![unclassifiedfailuresbtn](https://cloud.githubusercontent.com/assets/3660661/8658387/6787024c-2971-11e5-87fc-6805d11fcc03.jpg)

This is probably better than waiting for all the resultsets to load, since the test duration would be a bit shorter this way.

I can also add an ID for that button in treeherder. Do you think this is an ok approach @rbillings?

rbillings:
This sounds like a good approach. I'll add that I'm seeing failures locally that are failing differently- the test finds the unclassified failure, opens the job details, but is failing to open the log viewer button. This is a good place to start though.

tojonmz:
That's interesting. The only scenario I can think of where Logviewer wouldn't open with manual interaction, is if the Log is either not parsed, or in progress. We prevent that, since there's no log yet to view. The log parse status is visible on the selected job (in this case an unclassified failure) in the bottom of the Job details tab.

davehunt:
> I'll add that I'm seeing failures locally that are failing differently- the test finds the unclassified failure, opens the job details, but is failing to open the log viewer button.

I can replicate this, and I suspect we're not waiting long enough before we attempt to open the log viewer. It's passed 1/3 times for me.

rbillings:
[...]

PR #17, opened 2015-07-15
Added unclassified failure verification addresses Issue #14

rbillings:
#14

### DIFF
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -18,6 +18,7 @@ class ResultsetPage(Base):
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
     _result_status_locator = (By.ID, 'job-details-panel')
+    _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     @property
     def job_result_status(self):
@@ -33,6 +34,9 @@ class ResultsetPage(Base):
     def open_logviewer(self):
         WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._job_details_actionbar_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("l")
+
+    def return_unclassified_failure_count(self):
+        return self.selenium.find_element(*self.unclassified_failure_count_locator).get_attribute("value")
 
 class LogviewerPage(Base):
 
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -24,6 +24,10 @@ class ResultsetPage(Base):
     def job_result_status(self):
         return self.selenium.find_element(*self._job_result_status_locator).text
 
+    @property
+    def return_unclassified_failure_count(self):
+        return int(self.selenium.find_element(*self._unclassified_failure_count_locator).text)
+
     def go_to_page(self):
         self.open('')
 
@@ -34,9 +38,6 @@ class ResultsetPage(Base):
     def open_logviewer(self):
         WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._job_details_actionbar_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("l")
-
-    def return_unclassified_failure_count(self):
-        return self.selenium.find_element(*self.unclassified_failure_count_locator).int(element.text)
 
 class LogviewerPage(Base):
 
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -36,7 +36,7 @@ class ResultsetPage(Base):
         self.selenium.find_element(*self._resultset_locator).send_keys("l")
 
     def return_unclassified_failure_count(self):
-        return self.selenium.find_element(*self.unclassified_failure_count_locator).get_attribute("value")
+        return self.selenium.find_element(*self.unclassified_failure_count_locator).int(element.text)
 
 class LogviewerPage(Base):
 
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -16,22 +16,29 @@ class ResultsetPage(Base):
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
-    _result_status_locator = (By.ID, 'bottom-left-panel')
+    _result_status_locator = (By.ID, 'job-details-panel')
+    _unclassified_failure_count_locator = (By.CSS_SELECTOR, '.btn-unclassified-failures')
 
     @property
     def job_result_status(self):
         return self.selenium.find_element(*self._job_result_status_locator).text
 
+    @property
+    def verify_unclassified_failure_exists(self):
+        return self.selenium.find_element(*self.unclassified_failure_count_locator).text
+
     def go_to_page(self):
         self.open('')
 
     def open_next_unclassified_failure(self):
-        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._resultset_locator).is_displayed())
+        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._unclassified_failure_count_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("n")
 
     def open_logviewer(self):
         WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._result_status_locator).is_displayed())
-        self.selenium.find_element(*self._resultset_locator).send_keys("l")
+        self.selenium.find_element(*self._logviewer_button_locator).click()
+
+
 
 class LogviewerPage(Base):
 
--- a/pages/resultset.py
+++ b/pages/resultset.py
@@ -13,31 +13,26 @@ from pages.base import Base
 
 class ResultsetPage(Base):
 
+    _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
     _result_status_locator = (By.ID, 'job-details-panel')
-    _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     @property
     def job_result_status(self):
         return self.selenium.find_element(*self._job_result_status_locator).text
 
-    @property
-    def return_unclassified_failure_count(self):
-        return len(self.selenium.find_element(*self.unclassified_failure_count_locator))
-
     def go_to_page(self):
         self.open('')
 
     def open_next_unclassified_failure(self):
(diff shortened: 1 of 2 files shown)


pages/resultset.py
```diff
@@ -17,21 +17,27 @@ class ResultsetPage(Base):
...
 
     def open_next_unclassified_failure(self):
-        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._resultset_locator).is_displayed())
+        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._unclassified_failure_count_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("n")
 
     def open_logviewer(self):
         WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._result_status_locator).is_displayed())
-        self.selenium.find_element(*self._resultset_locator).send_keys("l")
+        self.selenium.find_element(*self._logviewer_button_locator).click()
```
davehunt: If we want to test the keyboard shortcuts then first waiting for the button to be displayed might work. If we don't care about shortcuts vs buttons, then this change is fine with me.

pages/resultset.py
```diff
@@ -17,21 +17,27 @@ class ResultsetPage(Base):
...
     _result_status_locator = (By.ID, 'job-details-panel')
+    _unclassified_failure_count_locator = (By.CSS_SELECTOR, '.btn-unclassified-failures')
 
     @property
     def job_result_status(self):
         return self.selenium.find_element(*self._job_result_status_locator).text
 
+    @property
+    def verify_unclassified_failure_exists(self):
+        return self.selenium.find_element(*self.unclassified_failure_count_locator).text
```
davehunt: Rather than send back a string that needs to be compared in the test, I would suggest a property that returns a count of the unclassified results displayed on the page. Having only a brief look, could we use the `btn-orange` or `btn-red` classes to identify unclassified results? If so, we could just return the length of `find_elements` and loop clicking the button until the return value is > 0. I would also be fine with having a `has_classified_jobs` in the page object that wraps this logic up.

pages/resultset.py
rbillings: The shortcuts are not the emphasis, the change was made in case it was a page load issue.

pages/resultset.py
rbillings: The reasoning behind this is that the count displays after the page loads, so it was a verification that the page is loaded and that an unclassified failure exists before proceeding with the test.  @tojonmz do you have an opinion on using btn-orange as a method to verify unclassified failures exist?

tojonmz: @rbillings @davehunt sure you'd want to check for `btn-orange`, `btn-red`, and `btn-purple` (infrastructure exception). They're all considered failures that potentially could appear. Whatever approach seems more robust sounds great to me.

davehunt: If Treeherder added a new class to indicate unclassified failures then that would probably be the most efficient way to proceed. I just tried excluding buttons with a class containing 'unclassified' but it also returns the green, blue and grey buttons (maybe others, too). There's no way to identify failures, let alone unclassified failures. Identifying them by their colour is not ideal and really not a good practice, see http://www.w3.org/QA/Tips/goodclassnames.

davehunt: Also, by returning the text we're returning (from example of production instance right now) "48 unclassified". This will resolve as True when treated as a boolean, but does not return an integer, and will likely never be "0". I think is would be "0 unclassified" at best.

tojonmz: Can we test this span to see if its value > 0 ? I could put an id on it to allow more robust selection if you like. Aiui it only updates after the DOM is populated, it's always a number (that could be converted from string if you wish in the test harness), and if the count button ever broke we'd indirectly be testing that also.
countspan

davehunt: Sounds good to me. We could use `.btn-unclassified-failures span` if we're confident that we're not going to break that, but an ID or class for that element would be ideal. I would suggest creating a property for the unclassified failures count that returns an integer based on this string, and directly checking that's greater than 0 in the tests.

tojonmz: > I would suggest creating a property for the unclassified failures count that returns an integer based on this string

That part I've excerpted above, is test-side, correct?

I'll add a suitable ID to the UI, as I'm pretty confident we'll only have one location in the page that will ever display the count. I'll let you know when that's landed and available for testing. :)

davehunt: Sorry @tojonmz, I don't understand your question.

tojonmz: Sorry @davehunt my question is, is this part below your/Rebecca's task, or mine?

> I would suggest creating a property for the unclassified failures count that returns an integer based on this string

davehunt: Ah, yes, we can take care of that. Let us know when the new ID is present. :)

tojonmz: Ok, almost there. I've merged the PR https://github.com/mozilla/treeherder/pull/877 but we still need to update stage.

tojonmz: Ok, stage is updated, any developer branches pushed to stage should also have this change :)

stage

rbillings:
@davehunt r?

tests/test_unclassified_job.py
```diff
@@ -21,6 +21,8 @@ def test_unclassified_failure(self, mozwebqa):
         # Open resultset page and search for next unclassified failure
         resultset_page = ResultsetPage(mozwebqa)
         resultset_page.go_to_page()
+        Assert.greater_equal("resultset_page.return_unclassified_failure_count", "1")
```
davehunt: This isn't doing what you're expecting. For comparing if something is greater or equal your values should be numbers. I would recommend `return_unclassified_failure_count` return an integer, and then you also need to change the value you're comparing it to an integer.

rbillings:
@davehunt - some of my former commits got omitted, so I re-added the failure count method with it returning an integer.

pages/resultset.py
```diff
@@ -30,9 +32,12 @@ def open_next_unclassified_failure(self):
         self.selenium.find_element(*self._resultset_locator).send_keys("n")
 
     def open_logviewer(self):
-        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._result_status_locator).is_displayed())
+        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._job_details_actionbar_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("l")
 
+    def return_unclassified_failure_count(self):
+        return self.selenium.find_element(*self.unclassified_failure_count_locator).get_attribute("value")
```
davehunt: This element doesn't appear to have an attribute named `value` so I suspect this will return `None`. I think you probably want the text content, coerced into an integer.

tests/test_unclassified_job.py
davehunt: We're still comparing with a string here.

tests/test_unclassified_job.py
```diff
@@ -34,6 +36,8 @@ def test_open_unclassified_failure_log(self, mozwebqa):
         # Open the job log and verify there is content
         resultset_page = ResultsetPage(mozwebqa)
         resultset_page.go_to_page()
+        Assert.not_equal("0","resultset_page.verify_unclassified_failure_exists")
```
davehunt: Again, we're comparing with a string here.

rbillings:
@davehunt r?

tests/test_unclassified_job.py
davehunt: This isn't returning the unclassified failure count from the page, it's checking that the string "resultset_page.return_unclassified_failure_count" is greater than the string "1". This happens to be true because 'f' has a higher value than '1'. See https://docs.python.org/2/tutorial/datastructures.html#comparing-sequences-and-other-types for details on how string comparisons are made.

To fix this, we need to call `resultset_page.return_unclassified_failure_count()` and compare it to `1` (not a string).

tests/test_unclassified_job.py
```diff
@@ -34,6 +36,8 @@ def test_open_unclassified_failure_log(self, mozwebqa):
         # Open the job log and verify there is content
         resultset_page = ResultsetPage(mozwebqa)
         resultset_page.go_to_page()
+        Assert.not_equal("0","resultset_page.return_unclassified_failure_count")
```
davehunt: Same issue as above.

pages/resultset.py
```diff
@@ -30,9 +32,12 @@ def open_next_unclassified_failure(self):
         self.selenium.find_element(*self._resultset_locator).send_keys("n")
 
     def open_logviewer(self):
-        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._result_status_locator).is_displayed())
+        WebDriverWait(self.selenium, self.timeout).until(lambda s: self.selenium.find_element(*self._job_details_actionbar_locator).is_displayed())
         self.selenium.find_element(*self._resultset_locator).send_keys("l")
 
+    def return_unclassified_failure_count(self):
+        return self.selenium.find_element(*self.unclassified_failure_count_locator).int(element.text)
```
davehunt: There's a typo in the locator variable name here. It needs a `_`.

davehunt: Also, this is trying to call an `int` method on the `WebElement` that doesn't exist. Also, you reference an `element` variable that doesn't exist. To cast a string to an integer you would need to wrap the entire call in `int()`. For example:

```python
```

davehunt: I would also recommend making this a property named `unclassified_failure_count`, then the return is implied an you can omit the `()` when calling it from the test.

tests/test_unclassified_job.py
davehunt: In fact, I think this should also check that the count is greater than 1 (or greater than zero).

rbillings:
@davehunt r?

pages/resultset.py
```diff
@@ -13,15 +13,21 @@
...
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
     _result_status_locator = (By.ID, 'job-details-panel')
+    _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     @property
     def job_result_status(self):
         return self.selenium.find_element(*self._job_result_status_locator).text
 
+    @property
+    def return_unclassified_failure_count(self):
```
davehunt: Nit: Remove `return_` from the property name as all properties should return a value.

tests/test_unclassified_job.py
```diff
@@ -34,6 +36,8 @@ def test_open_unclassified_failure_log(self, mozwebqa):
         # Open the job log and verify there is content
         resultset_page = ResultsetPage(mozwebqa)
         resultset_page.go_to_page()
+        Assert.not_equal(0, resultset_page.return_unclassified_failure_count)
```
davehunt: As it's a number, let's check this is greater than 0.

davehunt:
r+ with comments addressed

m8ttyB:
r+

win

tests/test_unclassified_job.py
```diff
@@ -34,6 +36,8 @@ def test_open_unclassified_failure_log(self, mozwebqa):
         # Open the job log and verify there is content
         resultset_page = ResultsetPage(mozwebqa)
         resultset_page.go_to_page()
+        Assert.greater_equal(1, resultset_page.unclassified_failure_count)
```
davehunt: This is checking that the unclassified failure count is less than or equal to 1, which is not what we want to do here.

rbillings:
Addressed last comment here https://github.com/mozilla/treeherder-tests/pull/21

Issue #4, opened 2015-05-27
Test Case - View a single resultset


tojon:
This test corresponds with Moztrap test case [15828](https://moztrap.mozilla.org/manage/case/15828/).

1. Load Treeherder and click on the 'Date' on the left side of any resultset.

> Only that resultset should load, with an accompanying URL param "&revision=(SHA)"

So we can check to ensure there is only one resultset in the DOM after the load, it contains a matching date, and in doing so validate the single resultset load worked.

The only scenario where this would be data dependent would be if the repo only contains one resultset to begin with - but with mozilla-inbound we can be confident that won't be the case.

davehunt:
Landed in https://github.com/mozilla/treeherder-tests/commit/ad19cac05d0c29e34e2edaee0e306558c3b3c750.

PR #27, opened 2015-11-10
Verifies able to select a single resultset

rbillings:
Fixes issue #4

### DIFF
--- a/pages/page.py
+++ b/pages/page.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/pages/page.py
+++ b/pages/page.py
@@ -19,6 +19,9 @@ class WebView(object):
     def _root(self):
         return self.selenium
 
+    def go_to_page(self):
+        self.open('')
+
     def find_element(self, locator):
         return self._root.find_element(*locator)
 
--- a/pages/page.py
+++ b/pages/page.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
@@ -20,9 +18,6 @@ class WebView(object):
     @property
     def _root(self):
         return self.selenium
-
-    def go_to_page(self):
-        self.open('')
 
     def find_element(self, locator):
         return self._root.find_element(*locator)
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -18,12 +18,6 @@ class TreeherderPage(Base):
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
-<<<<<<< 96097695cdecedfef212f6b55624f12978ef91d9:pages/treeherder.py
-    _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
-=======
-    _resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set .result-set-title-left > span a')
-    _result_status_locator = (By.ID, 'job-details-panel')
->>>>>>> Updated datestamp locator:pages/resultset.py
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     def wait_for_page_to_load(self):
@@ -44,7 +38,7 @@ class TreeherderPage(Base):
         return self.selenium.find_element(*self._first_resultset_datestamp_locator).text
 
     def open_next_unclassified_failure(self):
-        el = self.selenium.find_element(*self._resultset_datestamp_locator)
+        el = self.selenium.find_element(*self._first_resultset_datestamp_locator)
         Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
         el.send_keys('n')
         Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -17,6 +17,8 @@ class TreeherderPage(Base):
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
+    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
+    _single_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span:nth-child(2) > a:nth-child(1)')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     def wait_for_page_to_load(self):
@@ -35,6 +37,14 @@ class TreeherderPage(Base):
     @property
     def first_revision_date(self):
         return self.selenium.find_element(*self._first_resultset_datestamp_locator).text
+
+    @property
+    def revision_date(self):
+        return self.selenium.find_element(*self._single_resultset_datestamp_locator).text
+
+    @property
+    def results_count(self):
+        return len(self.selenium.find_elements(*self._results_locator))
 
     def open_next_unclassified_failure(self):
         el = self.selenium.find_element(*self._first_resultset_datestamp_locator)
@@ -58,7 +68,7 @@ class TreeherderPage(Base):
         Wait(self.selenium, self.timeout).until(
             EC.visibility_of_element_located(self._first_resultset_datestamp_locator))
         self.selenium.find_element(*self._first_resultset_datestamp_locator).click()
-        return SingleResultPage(self.base_url, self.selenium)
+        return self.selenium.find_element(*self._single_resultset_datestamp_locator)
 
 
 class LogviewerPage(Page):
@@ -73,22 +83,3 @@ class LogviewerPage(Page):
     @property
     def is_job_status_visible(self):
         return self.is_element_visible(self._job_header_locator)
-
-
-class SingleResultPage(Page):
-
-    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
-    _single_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span:nth-child(2) > a:nth-child(1)')
-
-    def __init__(self, base_url, selenium):
-        Page.__init__(self, base_url, selenium)
-        Wait(self.selenium, self.timeout).until(
-            EC.visibility_of_element_located(self._single_resultset_datestamp_locator))
-
-    @property
-    def revision_date(self):
-        return self.selenium.find_element(*self._single_resultset_datestamp_locator).text
-
-    @property
-    def results_count(self):
-        return len(self.selenium.find_elements(*self._results_locator))
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -18,7 +18,12 @@ class TreeherderPage(Base):
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
+<<<<<<< 96097695cdecedfef212f6b55624f12978ef91d9:pages/treeherder.py
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
+=======
+    _resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set .result-set-title-left > span a')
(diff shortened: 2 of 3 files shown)


rbillings:
@mozilla/web-qa-sorcerers r?

pages/resultset.py
```diff
@@ -12,6 +12,10 @@
 
 class ResultsetPage(Page):
 
+    _page_title = '[0] mozilla-inbound'
+
+    _first_resultset_locator = (By.CSS_SELECTOR, '.th-view-content > ng-view:nth-child(1) > div:nth-child(2) > div:nth-child(1) > span:nth-child(1) > span:nth-child(1) > span:nth-child(2) > a:nth-child(1)')
+    _first_resultset_number_locator = (By.CSS_SELECTOR, '.th-view-content > ng-view:nth-child(1) > div:nth-child(2) > div:nth-child(3) > div:nth-child(2) > span:nth-child(1) > ul:nth-child(1) > li:nth-child(2) > span:nth-child(1) > span:nth-child(1) > a:nth-child(1)')
```
davehunt: This isn't really a number, but a revision. Also, this is a rather complex locator. Looking at the DOM, I think this would be suitable: `.result-set .revision-holder a`.

Also, this locator isn't specific to the first match, so we don't need to prefix it with `first_`, we just only use the first match in the test/page object.

pages/resultset.py
```diff
@@ -12,6 +12,10 @@
 
 class ResultsetPage(Page):
 
+    _page_title = '[0] mozilla-inbound'
+
+    _first_resultset_locator = (By.CSS_SELECTOR, '.th-view-content > ng-view:nth-child(1) > div:nth-child(2) > div:nth-child(1) > span:nth-child(1) > span:nth-child(1) > span:nth-child(2) > a:nth-child(1)')
```
davehunt: I think we should name this `_resultset_link_locator` and the following would be a less fragile locator: `.result-set .result-set-title-left > span a`.

pages/resultset.py
```diff
@@ -32,6 +36,15 @@ def job_result_status(self):
     def unclassified_failure_count(self):
         return int(self.selenium.find_element(*self._unclassified_failure_count_locator).text)
 
+    def get_first_resultset_number(self):
```
davehunt: This should be a property, drop the `get` and change `number` to `revision`.

tests/test_view_single_result.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @pytest.mark.nondestructive
+    def test_open_single_result(self, base_url, selenium):
+        # Select single test result and verify the content
+        resultset_page = ResultsetPage(base_url, selenium).open()
+
+        resultsetnumber = resultset_page.get_first_resultset_number()
+        resultset_page.open_single_resultset()
+
+        assert resultset_page.get_url_current_page().endswith(resultsetnumber)
```
davehunt: I don't think we need to abstract the Selenium property for the current URL. This can simply be `assert selenium.current_url.endswith(revision)`

davehunt:
Sorry for the delay in review this @rbillings. You might want to consider creating a page region for the result set, as it's a repeating pattern. That way you could have a property for getting the revision and a method for viewing the result in isolation, and could reference the first result set via an index. Here's what the test might look like with the suggested changes.

```python
@pytest.mark.nondestructive
def test_open_single_result(self, base_url, selenium):
    resultset_page = ResultsetPage(base_url, selenium).open()
    revision = resultset_page.resultset[0].revision
    resultset_page.revision[0].view()
    assert selenium.current_url.endswith(revision)
```

rbillings:
Thanks for the suggestions @davehunt. I updated the names and locators. I also changed the test verification as I realized that some revisions have multiple elements, and the only unique factor for each revision is the date/time - which also better matches the original test case.

davehunt:
It looks like this needs a rebase, and also Travis seems to have had an issue.

pages/resultset.py
```diff
@@ -12,10 +12,12 @@
 
 class ResultsetPage(Page):
 
+    _page_title = '[0] mozilla-inbound'
+
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
-    _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
+    _resultset_locator = (By.CSS_SELECTOR, '.result-set .result-set-title-left > span a')
```
davehunt: It's not clear from the name that this is a locator for the date/timestamp now.

davehunt:
I still think a ResultSet page region makes sense here. I've updated my example to take into account the switch to the date/timestamp:

```python
@pytest.mark.nondestructive
def test_open_single_result(self, base_url, selenium):
    results_page = ResultsetPage(base_url, selenium).open()
    timestamp = results_page.results[0].timestamp
    single_result_page = results_page.results[0].view()
    assert 1 == len(single_result_page.results)
    assert timestamp == single_result_page.results[0]
```

rbillings:
@davehunt I absolutely agree that there should be a ResultSet page region. I did not make any changes in this PR for that as the other open PR contains many page region additions that would be needed here. After that PR is merged I will add the updates for creating a ResultSet page region. Thanks for the review!

rbillings:
The last update has a rebase, added page region and other fixes suggested. I did not add the suggested timestamp as the files have timestamps which are important- more than what time it is when the tests are run.

pages/page.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
```
davehunt: We don't need shebangs in files that are not going to be executed directly on the command line.

pages/page.py
```diff
@@ -19,6 +21,9 @@ def __init__(self, base_url, selenium, timeout=10, **kwargs):
     def _root(self):
         return self.selenium
 
+    def go_to_page(self):
+        self.open('')
+
```
davehunt: This is unused, and should be removed.

pages/treeherder.py
```diff
@@ -13,11 +13,11 @@
 
 class TreeherderPage(Base):
 
+    _first_resultset_locator = (By.CSS_SELECTOR, '.th-view-content > ng-view:nth-child(1) > div:nth-child(2) > div:nth-child(1) > span:nth-child(1) > span:nth-child(1) > span:nth-child(2) > a:nth-child(1)')
```
davehunt: This looks like a very fragile locator, and also doesn't appear to be used.

pages/treeherder.py
```diff
@@ -64,3 +74,22 @@ def __init__(self, base_url, selenium):
...
+
+class SingleResultPage(Page):
+
+    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
+    _single_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span:nth-child(2) > a:nth-child(1)')
+
+    def __init__(self, base_url, selenium):
+        Page.__init__(self, base_url, selenium)
+        Wait(self.selenium, self.timeout).until(
+            expected.window_with_title('mozilla-inbound'))
```
davehunt: This has a hard-coded branch name, which I suspect is not ideal.

davehunt:
> I did not add the suggested timestamp as the files have timestamps which are important- more than what time it is when the tests are run.

Sorry, I wasn't able to understand this, but perhaps it doesn't matter. The branch needs updating from master, and I've added a few comments.

rbillings:
closing/re-opening to trigger travis

rbillings:
@mozilla/web-qa-sorcerers r?

stephendonner:
I don't think my looking at this qualifies as a review, but I will comment that I got it to run locally without issue, and created our first ad-hoc job for Treeherder, here, and ran this pull through that, without issue (once I figured out the config): http://webqa-ci-staging1.qa.scl3.mozilla.com:8080/job/treeherder.adhoc/9/console

But again, I'm just getting back into the automation "game," so I've got very little skin :-(

pages/treeherder.py
```diff
@@ -64,3 +73,22 @@ def __init__(self, base_url, selenium):
     @property
     def is_job_status_visible(self):
         return self.is_element_visible(self._job_header_locator)
+
+
+class SingleResultPage(Page):
```
davehunt: Perhaps with the exception of the `__init__` this page object looks like it could be rolled up into `TreeherderPage`. The main results page shows > 1 result, and the single result page appears to be exactly the same in terms of structure, but only has one result. Therefore a property like `results_count` can apply to both. For the revision date, the existing `first_revision_date` would also work. Perhaps this can be addressed later, but I think we should at least improve the `_single_resultset_datestamp_locator` here, as it appears more fragile than `_first_resultset_datestamp_locator` even though they locate the same element.

tests/test_view_single_result.py
```diff
@@ -0,0 +1,19 @@
...
+import pytest
+
+from pages.treeherder import TreeherderPage
+
+
+class TestSingleResult:
+
+    @pytest.mark.nondestructive
+    def test_open_single_result(self, base_url, selenium):
+
```
davehunt: Nit: Remove this blank line.

pages/treeherder.py
rbillings: It's true that the format of the page does not change, but the url does- which is why a page region made sense to me. You cannot view the single result without an action. You could also say the same about the logviewer - it is part of the treeherder page but it only displays with an action. It's confusing to me.

Lastly I totally agree about the locators.

davehunt: Only the query string changes in the URL, the two pages most likely share the same view in the application. The log viewer has a very different DOM, and so I wouldn't consider it comparable to the result sets page.

You could have `SingleResultPage` extending `TreeherderPage` and then you'd only need to implement locators and functions that are specific to when there's a single result show in the page. That said, I suspect anything that's relevant to a single result shown is also relevant to multiple results.

rbillings:
I've addressed all previous comments. Ready for review.

pages/treeherder.py
```diff
@@ -13,11 +13,12 @@
...
 class TreeherderPage(Base):
 
+    _first_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set .result-set-title-left > span a')
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
-    _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
-    _result_status_locator = (By.ID, 'job-details-panel')
+    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
+    _single_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span:nth-child(2) > a:nth-child(1)')
```
davehunt: When there's only one resultset, our `_first_resultset_datestamp_locator` should still work.

pages/treeherder.py
```diff
@@ -33,8 +34,20 @@ def job_result_status(self):
     def unclassified_failure_count(self):
         return int(self.selenium.find_element(*self._unclassified_failure_count_locator).text)
 
+    @property
+    def first_revision_date(self):
+        return self.selenium.find_element(*self._first_resultset_datestamp_locator).text
+
+    @property
+    def revision_date(self):
+        return self.selenium.find_element(*self._single_resultset_datestamp_locator).text
```
davehunt: I think we can just call `first_revision_date`, it just so happens that there's only one result, so the first result is still accurate.

pages/treeherder.py
```diff
@@ -51,6 +64,12 @@ def open_perfherder_page(self):
         from perfherder import PerfherderPage
         return PerfherderPage(self.base_url, self.selenium).wait_for_page_to_load()
 
+    def open_single_resultset(self):
+        Wait(self.selenium, self.timeout).until(
+            EC.visibility_of_element_located(self._first_resultset_datestamp_locator))
+        self.selenium.find_element(*self._first_resultset_datestamp_locator).click()
+        return self.selenium.find_element(*self._single_resultset_datestamp_locator)
```
davehunt: You don't need to return anything here. The only thing that might be worth returning is a page object for the single result page, for which you could just return `self` as it's the same as the existing page (but with less results).

tests/test_view_single_result.py
```diff
@@ -0,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.treeherder import TreeherderPage
+
+
+class TestSingleResult:
```
davehunt: I'm going to raise this as a separate issue (and mark it as suitable for Outreachy applicants), but we really don't need test classes in pytest. In unittest they're required because you inherit from a base `TestClass` that gives you helpers for assertions, as well as other things. In pytest you can just define test functions.

davehunt:
Looking great @rbillings! Just a couple of comments, with those addressed this has my r+

rbillings:
@stephendonner - after checks are run, if you could review that would be great

rbillings:
@mozilla/web-qa-sorcerers r?

stephendonner:
Thx @rbillings - has addressed Dave's comments, runs for me locally, and also in the ad-hoc run: http://webqa-ci-staging1.qa.scl3.mozilla.com:8080/job/treeherder.adhoc/10/console

r+

stephendonner:
I think we can merge this after squashing these 8 commits, right?

davehunt:
Rebased, squashed, and merged: https://github.com/mozilla/treeherder-tests/commit/ad19cac05d0c29e34e2edaee0e306558c3b3c750

Issue #9, opened 2015-06-04
Test Case - Initial page load


tojon:
This test corresponds with Moztrap test case [15811](https://moztrap.mozilla.org/manage/case/15811/).

1. Load Treeherder from your specified instance: local, [stage](https://treeherder.allizom.org), or [production](https://treeherder.mozilla.org)
> Page loads displaying at least one resultset pushed to mozilla-inbound.

2. Treeherder logo > Perfherder
> Perfherder loads displaying its initial Graph page.

3. Perfherder logo > Treeherder
> Treeherder loads again, displaying resultsets per step 1.

rbillings:
Completed

PR #29, opened 2016-01-07
added a header region, created treeherder and perfherder pages, added…

rbillings:
… test to switch pages using header dropdown menu. Addresses Issue #9.

This test will ultimately be unnecessary as it only switches between Treeherder and Perfherder pages. However the Header region and the ability to switch pages will be used in other upcoming tests.

### DIFF
--- a/pages/base.py
+++ b/pages/base.py
@@ -11,13 +11,14 @@ class Base(Page):
 
     @property
     def header(self):
-        return Base.Header(self)
+        return self.Header(self)
 
     class Header(PageRegion):
 
+        _root_locator = (By.ID, 'th-global-navbar')
         _dropdown_menu_switch_page_locator = (By.CSS_SELECTOR, '.open ul > li a')
         _dropdown_menu_locator = (By.ID, 'th-logo')
 
         def switch_page_using_dropdown(self):
-            self.selenium.find_element(*self._dropdown_menu_locator).click()
-            self.selenium.find_element(*self._dropdown_menu_switch_page_locator).click()
+            self.find_element(self._dropdown_menu_locator).click()
+            self.find_element(self._dropdown_menu_switch_page_locator).click()
--- a/pages/base.py
+++ b/pages/base.py
@@ -1,0 +1,23 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+
+from pages.page import Page, PageRegion
+
+
+class Base(Page):
+
+    @property
+    def header(self):
+        return Base.Header(self)
+
+    class Header(PageRegion):
+
+        _dropdown_menu_switch_page_locator = (By.CSS_SELECTOR, '.open ul > li a')
+        _dropdown_menu_locator = (By.ID, 'th-logo')
+
+        def switch_page_using_dropdown(self):
+            self.selenium.find_element(*self._dropdown_menu_locator).click()
+            self.selenium.find_element(*self._dropdown_menu_switch_page_locator).click()
--- a/pages/perfherder.py
+++ b/pages/perfherder.py
@@ -1,0 +1,30 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait as Wait
+
+from pages.base import Base
+from pages.page import Page
+
+
+class PerfherderPage(Base):
+
+    _graph_chooser_locator = (By.ID, 'graph-chooser')
+
+    def wait_for_page_to_load(self):
+        Wait(self.selenium, self.timeout).until(
+            lambda s: self.is_graph_chooser_displayed)
+        return self
+
+    @property
+    def is_graph_chooser_displayed(self):
+        return self.is_element_visible(self._graph_chooser_locator)
+
+    def open_treeherder_page(self):
+        header = self.Header(self)
+        header.switch_page_using_dropdown()
+
+        from treeherder import TreeherderPage
+        return TreeherderPage(self.base_url, self.selenium).wait_for_page_to_load()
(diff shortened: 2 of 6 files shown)


pages/base.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Base(Page):
+
+    @property
+    def header(self):
+        return Base.Header(self)
+
+    class Header(PageRegion):
+
+        _dropdown_menu_switch_page_locator = (By.CSS_SELECTOR, '.open ul > li a')
```
davehunt: If you add `_root_locator = (By.ID, 'th-global-navbar')` then you'll be able to make sure any elements are beneath this element, rather than searching the entire document.

pages/base.py
```diff
@@ -0,0 +1,23 @@
...
+from selenium.webdriver.common.by import By
+
+from pages.page import Page, PageRegion
+
+
+class Base(Page):
+
+    @property
+    def header(self):
+        return Base.Header(self)
```
davehunt: Nit: You should be able to use `self.Header` here.

pages/perfherder.py
```diff
@@ -0,0 +1,30 @@
...
+            lambda s: self.is_graph_chooser_displayed)
+        return self
+
+    @property
+    def is_graph_chooser_displayed(self):
+        return self.is_element_visible(self._graph_chooser_locator)
+
+    def open_treeherder_page(self):
+        header = self.Header(self)
+        header.switch_page_using_dropdown()
```
davehunt: You should be able to just use `self.header.switch...` without setting a `header` variable.

pages/treeherder.py
```diff
@@ -44,6 +45,13 @@ def open_logviewer(self):
         self.selenium.find_element(*self._resultset_locator).send_keys('l')
         return LogviewerPage(self.base_url, self.selenium)
 
+    def open_perfherder_page(self):
+        header = self.Header(self)
+        header.switch_page_using_dropdown()
```
davehunt: As above, `self.header.switch...` should work here.

tests/test_unclassified_job.py
```diff
@@ -12,18 +12,18 @@ class TestUnclassifiedJobs:
     @pytest.mark.nondestructive
     def test_unclassified_failure(self, base_url, selenium):
         # Open resultset page and search for next unclassified failure
-        resultset_page = ResultsetPage(base_url, selenium).open()
-        assert resultset_page.unclassified_failure_count > 0
+        treeherder_page = TreeherderPage(base_url, selenium).open()
```
davehunt: I think you can get away with just calling this variable `page`. I've taken to doing this when there's only one page object in use.

tests/test_unclassified_job.py
```diff
@@ -12,18 +12,18 @@ class TestUnclassifiedJobs:
...
         assert teststatus in ['busted', 'testfailed', 'exception']
 
     @pytest.mark.nondestructive
     def test_open_unclassified_failure_log(self, base_url, selenium):
         # Open the job log and verify there is content
-        resultset_page = ResultsetPage(base_url, selenium).open()
-        assert resultset_page.unclassified_failure_count > 0
-        resultset_page.open_next_unclassified_failure()
-        logviewer_page = resultset_page.open_logviewer()
+        treeherder_page = TreeherderPage(base_url, selenium).open()
```
davehunt: As above, you could just call this `page`.

tests/test_treeherder_menu.py
```diff
@@ -0,0 +1,26 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.perfherder import PerfherderPage
+from pages.treeherder import TreeherderPage
+
+
+class TestTreeherderHeader:
```
davehunt: I've stopped using test classes as they don't really add anything unless there's more than one in a file. I would suggest removing this line, unindenting the rest of the file, and removing `self, ` from the test method signatures.

pages/base.py
```diff
@@ -0,0 +1,23 @@
...
+
+    class Header(PageRegion):
+
+        _dropdown_menu_switch_page_locator = (By.CSS_SELECTOR, '.open ul > li a')
+        _dropdown_menu_locator = (By.ID, 'th-logo')
+
+        def switch_page_using_dropdown(self):
+            self.selenium.find_element(*self._dropdown_menu_locator).click()
```
davehunt: If you grab these lines https://github.com/mozilla/bedrock/blob/22df5e90945f73cb54cfeda4a7204d5dc25c8f56/tests/pages/page.py#L18-L26 from bedrock's page.py then you'll be able to just use `self.find_element` and if a root locator is specified any lookups will be in the context of that root element. Note that you'll also be able to remove the `*`.

pages/perfherder.py
```diff
@@ -0,0 +1,30 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait as Wait
+
+from pages.base import Base
+from pages.page import Page
```
davehunt: Unused import.

tests/test_treeherder_menu.py
```diff
@@ -0,0 +1,26 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.perfherder import PerfherderPage
```
davehunt: Unused import.

davehunt:
Really nice improvements, @rbillings! I've made a few suggestions, and there are a couple of flake8 issues, but otherwise awesome work.

rbillings:
@davehunt - thanks for the suggestions. I added all except changing test_open_unclassified_failure_log to have "page" instead of "treeherder_page". The reason being that it goes between that and the logviewer page. It's true that the logviewer is in a way part of the treeherder page but right now it's being identified as a separate page.

davehunt:
r+ looks great!

rbillings:
@mozilla/web-qa-sorcerers r?

m8ttyB:
r+

PR #31, opened 2016-02-24
Update to pytest-selenium 1.2.0

davehunt:
There's a breaking change in v1.2.0, so I've released a candidate so that we can update our suites before the general release. @rbillings r?


rbillings:
Changes look good, tests pass. R+

PR #32, opened 2016-03-03
Updated login test

rbillings:
I'm not sure why I merged the original login commit as half of it was missing. This fixes it so the test passes.

### DIFF
--- a/pages/base.py
+++ b/pages/base.py
@@ -19,9 +19,9 @@ class Base(Page):
 
     class Header(PageRegion):
 
+        _root_locator = (By.ID, 'th-global-navbar')
         _login_locator = (By.CSS_SELECTOR, 'a.btn > span:nth-child(1)')
         _logout_icon_locator = (By.ID, 'logoutLabel')
-        _root_locator = (By.ID, 'th-global-navbar')
         _dropdown_menu_switch_page_locator = (By.CSS_SELECTOR, '.open ul > li a')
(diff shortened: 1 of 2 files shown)


pages/base.py
```diff
@@ -15,10 +19,27 @@ def header(self):
 
     class Header(PageRegion):
 
+        _login_locator = (By.CSS_SELECTOR, 'a.btn > span:nth-child(1)')
+        _logout_icon_locator = (By.ID, 'logoutLabel')
         _root_locator = (By.ID, 'th-global-navbar')
```
davehunt: Nit: We should put the `_root_locator` at the top.

pages/base.py
```diff
@@ -15,10 +19,27 @@ def header(self):
...
         _dropdown_menu_switch_page_locator = (By.CSS_SELECTOR, '.open ul > li a')
         _dropdown_menu_locator = (By.ID, 'th-logo')
 
+        @property
+        def is_user_logged_in(self):
+            return self.is_element_visible(self._logout_icon_locator)
+
+        def click_login(self):
+            self.selenium.find_element(*self._login_locator).click()
+            return not self.is_element_visible(self._login_locator)
```
davehunt: This function doesn't need to return a value.

davehunt:
This test doesn't appear to be running in Jenkins... perhaps it's being skipped due to the import error.

r+ with comments addressed.

rbillings:
Commits look happy, merging.

PR #33, opened 2016-03-08
Pinjob

### LINKED ISSUE #8: Test Case - Pin a job
This test corresponds with Moztrap test case [15814](https://moztrap.mozilla.org/manage/case/15814/).

1. Select a job, and click the 'pin' button in the lower navbar.
> The selected job is pinned.

2. Select another job, and hit [spacebar]
> The selected job is pinned.

3. Pinboard > Right hand menu dropdown > Clear all
> Both jobs are removed from the pinboard.

rbillings:
Fixes #8 - adds tests for pinning a job using spacebar, via the logviewer and clearing the pinboard.

### DIFF
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -10,6 +10,7 @@ from selenium.webdriver.common.keys import Keys
 import expected
 from pages.base import Base
 from pages.page import Page
+from pages.page import PageRegion
 
 
 class TreeherderPage(Base):
@@ -29,6 +30,7 @@ class TreeherderPage(Base):
     _result_status_locator = (By.ID, 'job-details-panel')
     _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
     _selected_job_title_locator = (By.CSS_SELECTOR, '.job-list .selected-job')
+    _result_set_locator = (By.CSS_SELECTOR, '#th-global-content .result-set')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     def wait_for_page_to_load(self):
@@ -37,8 +39,16 @@ class TreeherderPage(Base):
         return self
 
     @property
-    def job_result_status(self):
-        return self.selenium.find_element(*self._job_result_status_locator).text
+    def result(self):
+        return self.Result(self)
+
+    @property
+    def job_details(self):
+        return self.JobDetails(self)
+
+    @property
+    def pinboard(self):
+        return self.Pinboard(self)
 
     @property
     def unclassified_failure_count(self):
@@ -53,7 +63,7 @@ class TreeherderPage(Base):
         return len(self.selenium.find_elements(*self._results_locator))
 
     def open_next_unclassified_failure(self):
-        el = self.selenium.find_element(*self._first_resultset_datestamp_locator)
+        el = self.selenium.find_element(*self._result_set_locator)
         Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
         el.send_keys('n')
         Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
@@ -82,35 +92,80 @@ class TreeherderPage(Base):
         Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
         return self.selenium.find_element(*self._selected_job_title_locator).get_attribute('title')
 
-    def add_selected_job_to_pinboard(self):
-        el = self.selenium.find_element(*self._resultset_locator)
-        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
-        el.send_keys(Keys.SPACE)
-        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
 
-    @property
-    def is_pinboard_open(self):
-        return self.is_element_visible(self._pinboard_locator)
+    class Result(PageRegion):
 
-    @property
-    def pinned_job_title(self):
-        return self.selenium.find_element(*self._pinned_job_title_locator).get_attribute('title')
+        _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
+        _result_set_locator = (By.CSS_SELECTOR, '#th-global-content .result-set')
+        _selected_job_title_locator = (By.CSS_SELECTOR, '.job-list .selected-job')
 
-    @property
-    def pinboard_count(self):
-        return len(self.selenium.find_elements(*self._pinboard_count))
+        @property
+        def job_result_status(self):
+            return self.selenium.find_element(*self._job_result_status_locator).text
 
-    def pin_job_from_job_details(self):
-        self.select_next_job()
-        el = Wait(self.selenium, self.timeout).until(
-            EC.visibility_of_element_located(self._job_details_actionbar_locator))
-        el.find_element(*self._job_details_pin_job_locator).click()
-        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
+        @property
+        def result(self):
+            return [Result(self, root=el) for el in
+                self.find_elements(self._result_set_locator)]
 
-    def clear_pinboard(self):
-        self.selenium.find_element(*self._open_save_menu_locator).click()
-        self.selenium.find_element(*self._clear_all_menu_locator).click()
-        Wait(self.selenium, 20)
+        def pin_using_spacebar(self):
+            el = self.selenium.find_element(*self._result_set_locator)
+            Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+            el.send_keys(Keys.SPACE)
+            Wait(self.selenium, self.timeout).until(lambda s: self.pinboard.is_pinboard_open)
+
+        def select_next_job(self):
+            el = self.selenium.find_element(*self._result_set_locator)
+            Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+            el.send_keys(Keys.ARROW_RIGHT)
+            Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
+            return self.selenium.find_element(*self._selected_job_title_locator).get_attribute('title')
+
+    class JobDetails(PageRegion):
+
+        _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
+        _job_details_pin_job_locator = (By.CSS_SELECTOR, '#job-details-actionbar .nav .nav li:nth-child(1) a')
+        _logviewer_button_locator = (By.ID, 'logviewer-btn')
+        _result_status_locator = (By.ID, 'job-details-panel')
+
+        def open_logviewer(self):
+            Wait(self.selenium, self.timeout).until(
+                EC.visibility_of_element_located(self._job_details_actionbar_locator))
+            self.selenium.find_element(*self._result_set_locator).send_keys('l')
+            return LogviewerPage(self.base_url, self.selenium)
+
+        def pin_job(self):
+            el = Wait(self.selenium, self.timeout).until(
+                EC.visibility_of_element_located(self._job_details_actionbar_locator))
+            el.find_element(*self._job_details_pin_job_locator).click()
+            from pages.treeherder.pinboard import Pinboard
+            return Pinboard(self.base_url, self.selenium)
+
+
+    class Pinboard(PageRegion):
+
+        _clear_all_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-menu li:nth-child(4)')
+        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-toggle')
+        _pinboard_count = (By.CSS_SELECTOR, '#pinned-job-list')
+        _pinboard_locator = (By.ID, 'pinboard-panel')
+        _pinboard_remove_job_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job-close-btn')
+        _pinned_job_title_locator = (By.CSS_SELECTOR, '.pinned-job.selected-job')
+
+        @property
+        def pinned_job_title(self):
+            return self.selenium.find_element(*self._pinned_job_title_locator).get_attribute('title')
+
+        @property
+        def pins(self):
+            return len(self.selenium.find_elements(*self._pinboard_count))
+
+        @property
+        def is_pinboard_open(self):
+            return self.is_element_visible(self._pinboard_locator)
+
+        def clear_pinboard(self):
+            self.selenium.find_element(*self._open_save_menu_locator).click()
+            self.selenium.find_element(*self._clear_all_menu_locator).click()
 
 
 class LogviewerPage(Page):
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -67,7 +67,6 @@ class TreeherderPage(Base):
         el = self.selenium.find_element(*self._result_set_locator)
         Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
         el.send_keys(Keys.SPACE)
-        self.wait_for_page_to_load()
 
     def select_next_job(self):
         el = self.selenium.find_element(*self._result_set_locator)
@@ -99,7 +98,7 @@ class TreeherderPage(Base):
         _root_locator = (By.ID, 'pinboard-panel')
 
         _clear_all_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-menu li:nth-child(4)')
-        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls span.dropdown-toggle > span:nth-child(1)')
+        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .save-btn-dropdown')
         _pinboard_jobs_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job')
         _pinboard_remove_job_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job-close-btn')
         _selected_job_title_locator = (By.CSS_SELECTOR, '.pinned-job.selected-job')
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -14,16 +14,20 @@ from pages.page import Page
 
 class TreeherderPage(Base):
 
+    _clear_all_menu_locator = (By.CSS_SELECTOR, 'ul.dropdown-menu:nth-child(3) > li:nth-child(4) > a:nth-child(1)')
     _first_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set .result-set-title-left > span a')
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
-    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
+    _logviewer_pin_job_locator = (By.CSS_SELECTOR, 'a > span.glyphicon.glyphicon-pushpin')
     _next_job_title = (By.CSS_SELECTOR, '[class*="btn job-btn btn-dkgray"]')
+    _open_save_menu_locator = (By.CSS_SELECTOR, 'span.dropdown-toggle')
+    _pinboard_item_locator = (By.CSS_SELECTOR, 'span.btn-group')
     _pinboard_locator = (By.ID, 'pinboard-panel')
     _pinboard_remove_job_locator = (By.CSS_SELECTOR, 'i.fa-times')
     _resultset_locator = (By.CSS_SELECTOR, 'div.row.result-set')
     _result_status_locator = (By.ID, 'job-details-panel')
+    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     def wait_for_page_to_load(self):
@@ -83,12 +87,28 @@ class TreeherderPage(Base):
         el.send_keys(Keys.SPACE)
         Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
 
+    @property
     def is_pinboard_open(self):
         return self.is_element_visible(self._pinboard_locator)
 
     @property
     def pinned_job_title(self):
         return self.selenium.find_element(*self._next_job_title).get_attribute('title')
+
+    def pin_job_from_logviewer(self):
+        self.select_next_job()
+        el = Wait(self.selenium, self.timeout).until(
+            EC.visibility_of_element_located(self._job_details_actionbar_locator))
+        el.find_element(*self._logviewer_pin_job_locator).click()
+        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
+
+    def clear_pinboard(self):
+        self.selenium.find_element(*self._open_save_menu_locator).click()
+        self.selenium.find_element(*self._clear_all_menu_locator).click()
+
+    @property
+    def any_jobs_pinned(self):
+        return not self.is_element_visible(self._pinboard_item_locator)
 
 
 class LogviewerPage(Page):
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -93,13 +93,12 @@ class TreeherderPage(Base):
 
         def open_logviewer(self):
             self.wait_for_page_to_load
-            self.selenium.find_element(*self._job_details_panel_locator ).send_keys('l')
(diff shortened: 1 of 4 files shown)


tests/test_pin_job.py
```diff
@@ -0,0 +1,36 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from pages.treeherder import TreeherderPage
+
+
+class TestUnclassifiedJobs:
+
+    def test_pin_next_job(self, base_url, selenium):
+        # Open treeherder page, select next job and pin it
```
davehunt: Let's make these test descriptions docstrings:

```python
def test_pin_next_job(self, base_url, selenium):
    """Open treeherder page, select next job and pin it"""
```

pages/treeherder.py
```diff
@@ -13,9 +14,16 @@
 
 class TreeherderPage(Base):
 
+    _clear_all_menu_locator = (By.CSS_SELECTOR, 'ul.dropdown-menu:nth-child(3) > li:nth-child(4) > a:nth-child(1)')
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
+    _logviewer_pin_job_locator = (By.CSS_SELECTOR, 'a > span.glyphicon.glyphicon-pushpin')
+    _next_job_title = (By.CSS_SELECTOR, '[class*="btn job-btn btn-dkgray"]')
```
davehunt: This doesn't find anything for me. I suspect it's because of the `.btn-dkgray` which won't always find the next job. If you have an active job, and want to select the next job you could use `.selected-job + .job-btn`, but this won't work if there's no job already selected. In that case you'd want a locator for the first job on the page, something like just `.job-btn`.

pages/treeherder.py
```diff
@@ -51,6 +59,42 @@ def open_perfherder_page(self):
         from perfherder import PerfherderPage
         return PerfherderPage(self.base_url, self.selenium).wait_for_page_to_load()
 
+    def select_next_job(self):
+        el = self.selenium.find_element(*self._resultset_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.ARROW_RIGHT)
+        Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
+        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
```
davehunt: I would separate selecting the next job from retrieving the title of the next job, rather than returning it here. It's also a little unclear because would this be the next job after selecting the next job, or the current job after selecting the next job?

pages/treeherder.py
```diff
@@ -51,6 +59,42 @@ def open_perfherder_page(self):
         from perfherder import PerfherderPage
         return PerfherderPage(self.base_url, self.selenium).wait_for_page_to_load()
 
+    def select_next_job(self):
+        el = self.selenium.find_element(*self._resultset_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.ARROW_RIGHT)
+        Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
```
davehunt: This works when there isn't already a job selected, but if you called `select_next_job` twice, the second time this wait condition would evaluate as true potentially before the new job is selected.

pages/treeherder.py
```diff
@@ -51,6 +59,42 @@ def open_perfherder_page(self):
...
+        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
+
+    def add_selected_job_to_pinboard(self):
+        el = self.selenium.find_element(*self._resultset_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.SPACE)
+        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
```
davehunt: I think we'd want to wait for the new job to be pinned, as the pinboard may already be open. Perhaps we could wait for the number of pinned jobs to increase by one?

pages/treeherder.py
```diff
@@ -51,6 +59,42 @@ def open_perfherder_page(self):
...
+
+    @property
+    def is_pinboard_open(self):
+        return self.is_element_visible(self._pinboard_locator)
+
+    @property
+    def pinned_job_title(self):
+        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
```
davehunt: If this is meant to be the title for the selected pinned job, I think `.pinned-job.selected-job` would be the correct locator.

pages/treeherder.py
```diff
@@ -51,6 +59,42 @@ def open_perfherder_page(self):
...
+
+    def pin_job_from_logviewer(self):
```
davehunt: We've already used 'logviewer' to refer to the new window that displayed the job's logs. I would call this `job_details_panel` or similar.

pages/treeherder.py
```diff
@@ -51,6 +59,42 @@ def open_perfherder_page(self):
...
+        el.find_element(*self._logviewer_pin_job_locator).click()
+        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
+
+    def clear_pinboard(self):
+        self.selenium.find_element(*self._open_save_menu_locator).click()
+        self.selenium.find_element(*self._clear_all_menu_locator).click()
+
+    @property
+    def any_jobs_pinned(self):
+        return not self.is_element_visible(self._pinboard_item_locator)
```
davehunt: If we returned a total number of jobs in the pinboard, this could be used to assert that it's zero or non-zero.

pages/treeherder.py
```diff
@@ -13,9 +14,16 @@
 
 class TreeherderPage(Base):
 
+    _clear_all_menu_locator = (By.CSS_SELECTOR, 'ul.dropdown-menu:nth-child(3) > li:nth-child(4) > a:nth-child(1)')
```
davehunt: This looks like a fragile locator. First, I think we should anchor this to `pinboard-controls`. Ideally, we would get a class added to this `li` or `a`, but for now we'll have to rely on the position. Something like `#pinboard-controls .dropdown-menu li:nth-child(4)` would be better.

pages/treeherder.py
```diff
@@ -13,9 +14,16 @@
...
     _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
     _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span')
     _logviewer_button_locator = (By.ID, 'logviewer-btn')
+    _logviewer_pin_job_locator = (By.CSS_SELECTOR, 'a > span.glyphicon.glyphicon-pushpin')
```
davehunt: This is the job details panel rather than the logviewer, but the locator looks fragile. Again, it would be good to anchor it close to the element given the vagueness of the locator. I wonder if you could click the count? It might not work in Selenium if it's not the interactive element, but it might be worth a try: `#job-details-actionbar .pinned-job-count`

pages/treeherder.py
```diff
@@ -13,9 +14,16 @@
...
+    _next_job_title = (By.CSS_SELECTOR, '[class*="btn job-btn btn-dkgray"]')
+    _open_save_menu_locator = (By.CSS_SELECTOR, 'span.dropdown-toggle')
```
davehunt: Let's use `#pinboard-controls .dropdown-toggle` so this is never confused with another dropdown toggle in the page.

pages/treeherder.py
```diff
@@ -13,9 +14,16 @@
...
+    _pinboard_item_locator = (By.CSS_SELECTOR, 'span.btn-group')
```
davehunt: I suspect this is too vague. Looking at the DOM, perhaps `#pinned-job-list .pinned-job` would be better?

pages/treeherder.py
```diff
@@ -13,9 +14,16 @@
...
+    _pinboard_locator = (By.ID, 'pinboard-panel')
+    _pinboard_remove_job_locator = (By.CSS_SELECTOR, 'i.fa-times')
```
davehunt: Although this will return the first pinned job's remove element, I think `#pinned-job-list .pinned-job-close-btn` would be more suitable.

davehunt:
I think this is a decent approach, though I'd like to suggest an alternative. We could create some page regions for jobs, job details panel, and pinboard. This could simplify some of the page/region objects and allow for more readable test cases. Here's an example:

```python
def test_pin_job_from_logviewer(self, base_url, selenium):
    page = TreeherderPage(base_url, selenium).open()
    page.job[0].click()
    page.job_details.pin()
    assert 1 == len(page.pinboard.jobs)

def test_clear_pinboard(self, base_url, selenium):
    page = TreeherderPage(base_url, selenium).open()
    page.job[0].pin()
    assert 1 == len(page.pinboard.jobs)
    page.pinboard.clear()
    assert 0 == len(page.pinboard.jobs)
```

pages/treeherder.py
rbillings: That is true as the job details area remains open. I had no luck finding a way to ensure that the new job title is not the same as the previous job title that isn't super complicated. Is it worth checking the current title, moving to the next job and ensuring that the title has changed?

pages/treeherder.py
rbillings: Agreed- the locators have been changed to show "current" and "pinned" titles, which should be a lot more clear.

rbillings:
I pushed because I addressed many of the suggested changes and wanted to get those in. However:
** failure now because pinned jobs add time to the title attribute so they don’t match 
** pinboard_count is not counting correctly

rbillings:
Added page regions, made many of the suggested changes. I'm stuck on how to return a page region that is not within the current page region.

** failures treeherder.py 
line 101/ pin_job
line 74/pin_using_spacebar

pages/treeherder.py
```diff
@@ -52,6 +52,81 @@ def open_perfherder_page(self):
...
+            Wait(self.selenium, self.timeout).until(lambda s: self.pinboard.is_pinboard_open)
+
+        def select_next_job(self):
+            el = self.selenium.find_element(*self._result_set_locator)
+            Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+            el.send_keys(Keys.ARROW_RIGHT)
+            Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
+            return self.selenium.find_element(*self._selected_job_title_locator).get_attribute('title')
+
+    class JobDetails(PageRegion):
```
davehunt: This region should have a root locator of `#job-details-panel` and should use `self.find_element*` so that elements are located relative to this root element.

davehunt: Hmm.. I see that all locators are already using IDs, and you're using Selenium's expected conditions, which locate from the document root, so perhaps this can be left as it is.

pages/treeherder.py
```diff
@@ -52,6 +52,81 @@ def open_perfherder_page(self):
...
+
+        _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
+        _job_details_pin_job_locator = (By.CSS_SELECTOR, '#job-details-actionbar .nav .nav li:nth-child(1) a')
+        _logviewer_button_locator = (By.ID, 'logviewer-btn')
+        _result_status_locator = (By.ID, 'job-details-panel')
+
+        def open_logviewer(self):
+            Wait(self.selenium, self.timeout).until(
+                EC.visibility_of_element_located(self._job_details_actionbar_locator))
```
davehunt: This wait is in both methods, perhaps we can move it to an `__init__` instead.

pages/treeherder.py
```diff
@@ -52,6 +52,81 @@ def open_perfherder_page(self):
...
+            self.selenium.find_element(*self._result_set_locator).send_keys('l')
+            return LogviewerPage(self.base_url, self.selenium)
+
+        def pin_job(self):
+            el = Wait(self.selenium, self.timeout).until(
+                EC.visibility_of_element_located(self._job_details_actionbar_locator))
+            el.find_element(*self._job_details_pin_job_locator).click()
+            from pages.treeherder.pinboard import Pinboard
+            return Pinboard(self.base_url, self.selenium)
```
davehunt: PageRegions take the parent page as the argument, and extract the base URL and Selenium object from there.

pages/treeherder.py
```diff
@@ -52,6 +52,81 @@ def open_perfherder_page(self):
...
+
+
+    class Pinboard(PageRegion):
```
davehunt: It's a good practice for page regions to define a root locator. In this case I believe it would be:

```python
_root_locator = (By.ID, 'pinboard-panel')
```

pages/treeherder.py
```diff
@@ -52,6 +52,81 @@ def open_perfherder_page(self):
...
+
+        _clear_all_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-menu li:nth-child(4)')
+        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-toggle')
+        _pinboard_count = (By.CSS_SELECTOR, '#pinned-job-list')
```
davehunt: I would call this locator `_pinned_jobs_locator` and it should be `#pinned-job-list .pinned-job`

tests/test_pin_job.py
```diff
@@ -0,0 +1,36 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from pages.treeherder import TreeherderPage
+
+class TestUnclassifiedJobs:
+
+    def test_pin_next_job(self, base_url, selenium):
+        """ Open treeherder page, select next job and pin it"""
+        page = TreeherderPage(base_url, selenium).open()
+        current_job_title = page.result.select_next_job()
```
davehunt: The page is split into several result sets, each has one or more jobs. I think the regions in the page object model should reflect that, so the test would be page.result_set[0].jobs[0].click() to select the first job of the first result set. I know this is a little different from my previous comment, as I'm taking into account multiple result sets. I think a `select_next_job` function is still valuable, but it feels like it should be a top level function in the `TreeherderPage` rather than within a page region.

tests/test_pin_job.py
```diff
@@ -0,0 +1,36 @@
...
+        original_count = page.pinboard.pins
```
davehunt: You should be able to do an `assert 0 == len(page.pinboard.pinned_jobs)`

tests/test_pin_job.py
```diff
@@ -0,0 +1,36 @@
...
+
+        assert 1 == len(page.pinboard.pins)
+        assert current_job_title == page.pinboard.pinned_job_title
+
+    def test_pin_job_from_job_details(self, base_url, selenium):
+        """ Open treeherder page, select next job, pin it by the logviewer icon"""
+        page = TreeherderPage(base_url, selenium).open()
+
+        next_job_title = page.result.select_next_job()
+        page.job_details.pin_job()
```
davehunt: This looks great. I think we should add checks for the pinned job count before/after pinning.

davehunt:
Looking really good. Let me know if you want me to talk through any of my comments.

rbillings:
I've updated the page regions, fixed the tests & rebased from master. Ready for review! @mozilla/web-qa-sorcerers

rbillings:
Would someone please give this a r? I'm looking at you @mozilla/web-qa-sorcerers

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+        el = self.selenium.find_element(*self._result_set_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.ARROW_RIGHT)
+        Wait(self.selenium, self.timeout).until(lambda s: self.result.job_result_status)
+        return self.selenium.find_element(*self._selected_job_title_locator).get_attribute('title')
+
+    class JobDetails(PageRegion):
+
+        _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
+        _job_details_pin_job_locator = (By.CSS_SELECTOR, '#job-details-actionbar .nav .nav li:nth-child(1) a')
```
davehunt: Nit: These two locators don't need the `_job_details` prefix now that they're within a `JobDetails` region.

davehunt: This locator looks fragile, is there a better way to locate this element? I see the `span` beneath the `a` has a class of `glyphicon-pushpin` but I suspect the click only works on the `a` element. Perhaps something we could file a bug against Treeherder for adding a class to the `a` element?

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+        _logviewer_button_locator = (By.ID, 'logviewer-btn')
+        _result_status_locator = (By.ID, 'job-details-panel')
+
+        def wait_for_page_to_load(self):
+            body = self.find_element(By.TAG_NAME, 'body')
+            self.wait.until(lambda s: 'loaded' in body.get_attribute('JobDetails'))
```
davehunt: I don't see an attribute of `JobDetails` on the `body` element. Also, as this is a region and not a page, I don't think this method is ever called.

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+        el.send_keys(Keys.ARROW_RIGHT)
+        Wait(self.selenium, self.timeout).until(lambda s: self.result.job_result_status)
+        return self.selenium.find_element(*self._selected_job_title_locator).get_attribute('title')
+
+    class JobDetails(PageRegion):
+
+        _job_details_actionbar_locator = (By.ID, 'job-details-actionbar')
+        _job_details_pin_job_locator = (By.CSS_SELECTOR, '#job-details-actionbar .nav .nav li:nth-child(1) a')
+        _logviewer_button_locator = (By.ID, 'logviewer-btn')
+        _result_status_locator = (By.ID, 'job-details-panel')
```
davehunt: The name is misleading as this locates the panel element.

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+
+        def wait_for_page_to_load(self):
+            body = self.find_element(By.TAG_NAME, 'body')
+            self.wait.until(lambda s: 'loaded' in body.get_attribute('JobDetails'))
+
+        def open_logviewer(self):
+            self.wait_for_page_to_load
```
davehunt: Ah, so this is where the above method is being called. Without the `()` the method will return as an object rather than execute.

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+        _root_locator = (By.ID, 'pinboard-panel')
+
+        _clear_all_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-menu li:nth-child(4)')
+        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls span.dropdown-toggle > span:nth-child(1)')
+        _pinboard_jobs_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job')
+        _pinboard_remove_job_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job-close-btn')
+        _pinned_job_title_locator = (By.CSS_SELECTOR, '.pinned-job.selected-job')
+
+        @property
+        def pinned_job_title(self):
```
davehunt: The property and locator should indicate this is the selected job, rather than the pinned job.

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+
+        def clear_pinboard(self):
+            self.selenium.find_element(*self._open_save_menu_locator).click()
+            self.selenium.find_element(*self._clear_all_menu_locator).click()
+
+    class Result(PageRegion):
+
+        _root_locator = (By.ID, 'info-panel-content')
+
+        _job_result_status_locator = (By.CSS_SELECTOR, '#result-status-pane > div:nth-child(1) > span:nth-child(2)')
```
davehunt: I think this belongs in `JobDetails`, though I may be misunderstanding the `Result` region.

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+        _result_set_locator = (By.CSS_SELECTOR, '#th-global-content .result-set')
```
davehunt: This element is not under the root locator. I think `pin_using_spacebar` could either be moved to `TreeherderPage` or we could create a `Job` class within a `ResultSet` class with a `pin` method.

pages/treeherder.py
```diff
@@ -59,10 +52,79 @@ def open_perfherder_page(self):
...
+
+        def wait_for_page_to_load(self):
+            body = self.find_element(By.TAG_NAME, 'body')
+            self.wait.until(lambda s: 'loaded' in body.get_attribute('Result'))
```
davehunt: I don't think this method is ever called.

tests/test_treeherder_menu.py
```diff
@@ -9,15 +9,16 @@
 
 @pytest.mark.nondestructive
 def test_treeherder_dropdown(base_url, selenium):
-    # Switch between Treeherder and Perfherder using header dropdown
+    """ Switch between Treeherder and Perfherder using header dropdown """
     treeherder_page = TreeherderPage(base_url, selenium).open()
-    # Verify Treeherder loads
+
+    """ Verify Treeherder loads """
```
davehunt: Nit: The `"""` style is for docstrings, and should only be used in the first line of a method.
For other comments it's fine to use `#`.

pages/treeherder.py
rbillings: I definitely tried to make the glyphicon-pushpin to work before getting this locator. I'll file a bug for adding a class to the a element.

pages/treeherder.py
rbillings: So this is confusing- I want the open_logviewer method to execute. Are you suggesting it change?

pages/treeherder.py
rbillings: Good catch as the jobs are on the pinboard but are not technically saved yet.

pages/treeherder.py
rbillings: The intention was for JobDetails to be the specific job results pane and the Result region to be for jobs listed within a result-set. After re-jiggering I realized I created the Result region for no real reason- so this should be cleared up with my patch.

rbillings:
Comments added, comments addressed, removed an unneeded page region. Ready for another r?

pages/treeherder.py
davehunt: `self.wait_for_page_to_load` will do nothing, but `self.wait_for_page_to_load()` would execute the code contained in the `wait_for_page_to_load` method.

davehunt: Also, this is a region so we should call this method 'wait_for_region_to_load'. I suspect you perhaps don't even need it as the wait looks like it will never return true, and you're not currently calling it, yet your tests are likely passing?

pages/treeherder.py
```diff
@@ -64,6 +63,69 @@ def open_single_resultset(self):
...
+            body = self.find_element(By.TAG_NAME, 'body')
+            self.wait.until(lambda s: 'loaded' in body.get_attribute('JobDetails'))
+
+        def open_logviewer(self):
+            self.wait_for_page_to_load
+            self.selenium.find_element(*self._job_details_panel_locator).send_keys('l')
+            return LogviewerPage(self.base_url, self.selenium)
+
+        def pin_job(self):
+            self.wait_for_page_to_load
```
davehunt: As above, this is doing nothing, perhaps it's not needed at all?

rbillings:
I've updated to remove unused waits. R+?

pages/treeherder.py
```diff
@@ -64,6 +63,63 @@ def open_single_resultset(self):
...
+        @property
+        def pins(self):
+            return len(self.selenium.find_elements(*self._pinboard_jobs_locator))
+
+        @property
+        def is_pinboard_open(self):
+            return self.is_element_visible(self._root_locator)
+
+        def clear_pinboard(self):
+            self.selenium.find_element(*self._open_save_menu_locator).click()
```
davehunt: This is failing for me with element not found. Changing the locator to `#pinboard-controls .save-btn-dropdown` fixes it locally.

pages/treeherder.py
```diff
@@ -64,6 +63,63 @@ def open_single_resultset(self):
             EC.visibility_of_element_located(self._first_resultset_datestamp_locator))
         self.selenium.find_element(*self._first_resultset_datestamp_locator).click()
 
+    def pin_using_spacebar(self):
+        el = self.selenium.find_element(*self._result_set_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.SPACE)
+        self.wait_for_page_to_load()
```
davehunt: I don't think we need this as we're not causing a page load.

davehunt:
One test failure due to a locator, and one potentially unnecessary wait. Otherwise this looks good.

Issue #6, opened 2015-05-29
Test Case - Pin all visible jobs in resultset


tojon:
This test corresponds with Moztrap test case [15823](https://moztrap.mozilla.org/manage/case/15823/).

1. Select a job in the first resultset loaded.
> Confirm the pinboard is empty.

2. Click on the *Pin all visible jobs in resultset* pin-icon in the right hand side of that resultset bar.
>  Confirm the pinboard contains > 0 and < 500 jobs pinned. nb. 500 is our current limit, if hit a momentary notification warning will appear onscreen during the test.

3. Click in the pinboard on the extreme right hand drop down menu, and select 'Clear all'.
> Confirm the pinboard is empty.

PR #41, opened 2016-04-06
Pin all jobs

rbillings:
Create a test to pin all jobs in a test result, addresses issue #6

### DIFF
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -164,6 +164,26 @@ class TreeherderPage(Base):
             Wait(self.selenium, self.timeout).until(lambda _: el.get_attribute('aria-expanded') == 'true')
             self.selenium.find_element(*self._clear_all_menu_locator).click()
 
+    def select_next_job(self):
+        el = self.selenium.find_element(*self._resultset_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.ARROW_RIGHT)
+        Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
+        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
+
+    def add_selected_job_to_pinboard(self):
+        el = self.selenium.find_element(*self._resultset_locator)
+        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
+        el.send_keys(Keys.SPACE)
+        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
+
+    def is_pinboard_open(self):
+        return self.is_element_visible(self._pinboard_locator)
+
+    @property
+    def pinned_job_title(self):
+        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
+
         class Job(PageRegion):
 
             @property
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -66,8 +66,8 @@ class TreeherderPage(Base):
         from perfherder import PerfherderPage
         return PerfherderPage(self.base_url, self.selenium).wait_for_page_to_load()
 
-    def open_repos_menu(self):
-        self.selenium.find_element(*self._repos_menu_locator).click()
+    def pin_all_jobs(self):
+        self.selenium.find_element(*self._pin_all_jobs_locator).click()
 
     def pin_using_spacebar(self):
         el = self.selenium.find_element(*self._result_sets_locator)
@@ -163,16 +163,6 @@ class TreeherderPage(Base):
             el.click()
             Wait(self.selenium, self.timeout).until(lambda _: el.get_attribute('aria-expanded') == 'true')
             self.selenium.find_element(*self._clear_all_menu_locator).click()
-
-    def select_next_job(self):
-        el = self.selenium.find_element(*self._resultset_locator)
-        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
-        el.send_keys(Keys.ARROW_RIGHT)
-        Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
-        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
-
-    def pinned_job_title(self):
-        return self.selenium.find_element(*self._next_job_title).get_attribute('title')
 
         class Job(PageRegion):
 
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -171,16 +171,6 @@ class TreeherderPage(Base):
         Wait(self.selenium, self.timeout).until(lambda s: self.job_result_status)
         return self.selenium.find_element(*self._next_job_title).get_attribute('title')
 
-    def add_selected_job_to_pinboard(self):
-        el = self.selenium.find_element(*self._resultset_locator)
-        Wait(self.selenium, self.timeout).until(EC.visibility_of(el))
-        el.send_keys(Keys.SPACE)
-        Wait(self.selenium, self.timeout).until(lambda s: self.is_pinboard_open)
-
-    def is_pinboard_open(self):
-        return self.is_element_visible(self._pinboard_locator)
-
-    @property
     def pinned_job_title(self):
         return self.selenium.find_element(*self._next_job_title).get_attribute('title')
 
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -18,11 +18,7 @@ from pages.page import PageRegion
 
 class TreeherderPage(Base):
 
-    _active_watched_repo_locator = (By.CSS_SELECTOR, '#watched-repo-navbar button.active')
-    _mozilla_central_repo_locator = (By.CSS_SELECTOR, '#th-global-navbar-top a[href*="mozilla-central"]')
-    _repos_menu_locator = (By.ID, 'repoLabel')
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
-    _unchecked_repos_links_locator = (By.CSS_SELECTOR, '#repoLabel + .dropdown-menu .dropdown-checkbox:not([checked]) + .dropdown-link')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
 
     def wait_for_page_to_load(self):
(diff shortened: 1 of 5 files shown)


pages/treeherder.py
```diff
@@ -17,6 +17,7 @@ class TreeherderPage(Base):
 
     _result_set_locator = (By.CSS_SELECTOR, '#th-global-content .result-set')
     _selected_job_title_locator = (By.CSS_SELECTOR, '.job-list .selected-job')
+    _pin_all_jobs_locator = (By.CSS_SELECTOR, '.th-view-content > ng-view:nth-child(1) > div:nth-child(2) > div:nth-child(1) > span:nth-child(3) > span:nth-child(2) > span:nth-child(1)')
```
davehunt: This looks like a fragile locator. I think `.result-set .glyphicon-pushpin` would work just as well. I actually think it would be worth creating a `ResultSet` page region, which would allow us to target any result set's children elements.

tests/test_pin_job.py
```diff
@@ -36,3 +36,9 @@ def test_clear_pinboard(self, base_url, selenium):
         page.pinboard.clear_pinboard()
         assert page.pinboard.is_pinboard_open
         assert 0 == page.pinboard.pins
+
+    def test_pin_all_jobs(self, base_url, selenium):
+        """ Open treeherder page, pin all jobs, confirm no more than 500 pins in pinboard"""
+        page = TreeherderPage(base_url, selenium).open()
+        page.pin_all_jobs()
+        assert 500 >= page.pinboard.pins
```
davehunt: We can't guarantee there's going to be > 500 jobs. As mentioned in my previous comment, I think a `ResultSet` region would allow us to do more here. For example, the test could look like this:

```python
page.result_sets[0].pin_all_jobs()
assert len(page.result_sets[0].jobs) == len(page.pinboard.pins)
```

rbillings:
Fixed conflicts, updated test to match test_pin_jobs. R?

davehunt:
This is blocked by #33 and will need to be rebased once that lands. If we land #42 then we may also be able to match the number of jobs expected to be pinned.

rbillings:
Updated after #33 and #42 were merged. I would expect the "Pin all" test to totally pass or fail, which is why I didn't think counting the number of jobs was needed. Ready for R?

pages/treeherder.py
```diff
@@ -15,7 +15,11 @@
 
 class TreeherderPage(Base):
 
+    _first_resultset_datestamp_locator = (By.CSS_SELECTOR, '.result-set .result-set-title-left > span a')
```
davehunt: Unused locator

pages/treeherder.py
```diff
@@ -15,7 +15,11 @@
...
+    _pin_all_jobs_locator = (By.CSS_SELECTOR, '.result-set .glyphicon-pushpin')
```
davehunt: This should be within `ResultSet` region, as there's a pin all icon for each result set

pages/treeherder.py
```diff
@@ -15,7 +15,11 @@
...
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
+    _results_locator = (By.CSS_SELECTOR, '.result-set-bar')
+    _selected_job_title_locator = (By.CSS_SELECTOR, '.job-list .selected-job')
```
davehunt: These two locators are unused

pages/treeherder.py
```diff
@@ -24,6 +28,10 @@ def wait_for_page_to_load(self):
         return self
 
     @property
+    def first_revision_date(self):
```
davehunt: This is not used

pages/treeherder.py
```diff
@@ -51,6 +59,9 @@ def open_perfherder_page(self):
         from perfherder import PerfherderPage
         return PerfherderPage(self.base_url, self.selenium).wait_for_page_to_load()
 
+    def pin_all_jobs(self):
+        self.selenium.find_element(*self._pin_all_jobs_locator).click()
```
davehunt: We should move this to the `ResultSet` region

tests/test_pin_job.py
```diff
@@ -18,7 +18,7 @@ def test_pin_job(self, base_url, selenium):
         assert job.symbol == page.pinboard.selected_job.symbol
 
     def test_pin_job_from_job_details(self, base_url, selenium):
-        # Open treeherder page, select first job, pin it by the pin icon
+        """ Open treeherder page, select next job, pin it by the logviewer icon"""
```
davehunt: I'm not sure I understand this change. The test selects the first job (next seems ambiguous), and does not use a logview icon at all.

davehunt: Nit: Remove the whitespace after the opening `"""`

tests/test_pin_job.py
```diff
@@ -36,3 +36,9 @@ def test_clear_pinboard(self, base_url, selenium):
         page.pinboard.clear_pinboard()
         assert page.pinboard.is_pinboard_open
         assert 0 == len(page.pinboard.jobs)
+
+    def test_pin_all_jobs(self, base_url, selenium):
+        """ Open treeherder page, pin all jobs, confirm no more than 500 pins in pinboard"""
```
davehunt: Nit: Remove the whitespace after the opening """

tests/test_treeherder_menu.py
```diff
@@ -9,16 +9,16 @@
 
 @pytest.mark.nondestructive
 def test_treeherder_dropdown(base_url, selenium):
-    # Switch between Treeherder and Perfherder using header dropdown
+    """ Switch between Treeherder and Perfherder using header dropdown """
```
davehunt: Nit: Remove the whitespace after the opening """

tests/test_treeherder_menu.py
```diff
@@ -9,16 +9,16 @@
...
     treeherder_page = TreeherderPage(base_url, selenium).open()
 
-    # Verify Treeherder loads
+    """ Verify Treeherder loads """
```
davehunt: This should be an inline comment, not a docstring

tests/test_treeherder_menu.py
```diff
@@ -9,16 +9,16 @@
...
     assert treeherder_page.unclassified_failure_count > 0
 
-    # Switch to Perfherder page
+    """ Switch to Perfherder page """
```
davehunt: This should be an inline comment, not a docstring

tests/test_treeherder_menu.py
```diff
@@ -9,16 +9,16 @@
...
     perfherder_page = treeherder_page.open_perfherder_page()
     assert perfherder_page.is_graph_chooser_displayed
 
-    # Return to Treeherder page
+    """ Return to Treeherder page """
```
davehunt: This should be an inline comment, not a docstring

tests/test_unclassified_job.py
```diff
@@ -11,7 +11,7 @@ class TestUnclassifiedJobs:
 
     @pytest.mark.nondestructive
     def test_unclassified_failure(self, base_url, selenium):
-        # Open resultset page and search for next unclassified failure
+        """ Open resultset page and search for next unclassified failure"""
```
davehunt: Nit: Remove the whitespace after the opening """

tests/test_unclassified_job.py
```diff
@@ -21,7 +21,7 @@ def test_unclassified_failure(self, base_url, selenium):
 
     @pytest.mark.nondestructive
     def test_open_unclassified_failure_log(self, base_url, selenium):
-        # Open the job log and verify there is content
+        """ Open the job log and verify there is content """
```
davehunt: Nit: Remove the whitespace after the opening """

rbillings:
Updated all the docstrings, moved pin all to the result set page region. R?

pages/treeherder.py
```diff
@@ -61,6 +61,7 @@ class ResultSet(PageRegion):
 
         _datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span a')
         _jobs_locator = (By.CLASS_NAME, 'job-btn')
+        _pin_all_jobs_locator = (By.CSS_SELECTOR, '.result-set .glyphicon-pushpin')
```
davehunt: We should be able to just use `(By.CLASS_NAME, '.result-set .glyphicon-pushpin')` and find from the context of the root element.

pages/treeherder.py
```diff
@@ -70,6 +71,9 @@ def datestamp(self):
         def jobs(self):
             return [self.Job(self.page, root=el) for el in self.find_elements(self._jobs_locator)]
 
+        def pin_all_jobs(self):
+            self.selenium.find_element(*self._pin_all_jobs_locator).click()
```
davehunt: This will find the element from the document (the first one in the page) instead of from the context of the page region. Change it to `self.find_element` and it will find from the root element.

tests/test_treeherder_menu.py
davehunt: There's still whitespace here, now also before the closing `"""`

tests/test_treeherder_menu.py
davehunt: This is still a docstring

tests/test_treeherder_menu.py
davehunt: This is still a docstring

tests/test_treeherder_menu.py
davehunt: This is still a docstring

tests/test_unclassified_job.py
davehunt: Still has whitespace

tests/test_unclassified_job.py
davehunt: There's still whitespace here, now also before the closing `"""`

pages/treeherder.py
rbillings: Changing this to By.CLASS_NAME generates this error locally:
E       InvalidSelectorException: Message: The given selector .result-set .glyphicon-pushpin is either invalid or does not result in a WebElement. The following error occurred:
E       InvalidSelectorError: Compound class names not permitted

Were you able to get it to pass this way?

davehunt: Sorry, error on my part... if it's just a class name you don't want a CSS selector:

```python
_pin_all_jobs_locator = (By.CLASS_NAME, 'glyphicon-pushpin')
```

rbillings:
Updated to fix comments and merge conflicts. R?

stephendonner:
Running this locally, I sometimes (2/5 times or so?) get a failure in test_clear_pinboard.  Here's the full stack: https://pastebin.mozilla.org/8867755

davehunt:
> Running this locally, I sometimes (2/5 times or so?) get a failure in test_clear_pinboard. Here's the full stack: https://pastebin.mozilla.org/8867755

This is unrelated to the changes in this patch. My best guess is that the job details and pinboard panels are still moving around when we attempt to interact with the dropdown menu item. This exception means that another element is obstructing the target click location. It most probably means we need to add more explicit waits around our interactions. I think we should file this as a separate issue.

davehunt:
r+ merging and squashing

PR #44, opened 2016-04-14
Wait for page to load after switching repository

davehunt:
This should solve an intermittent failure when switching repository.

@rbillings r?


rbillings:
Change looks good, and helps the tests pass. R+

PR #55, opened 2016-04-30
Logtitle

rbillings:
Recently the logviewer page removed the portion of the title that verification was based on, this updates the check for a web element.


stephendonner:
Thx; this LGTM, and is cleaner than before, so r+.  It fixes my persistent failure in 09:19:53 [gw11] FAILED tests/test_unclassified_job.py::test_open_unclassified_failure_log

Issue #67, opened 2016-05-11
Test: Add new jobs


rbillings:
Create test to add new jobs to a resultset.

1) Open Treeherder
2) Select any resultset, open dropdown on far right
3) Select "add new jobs" option

Verify new  “job-like-items” with the state of “runnable” are added to the resultset, they display in a golden color. 

4) Select same resultset, open dropdown on far right, select "Hide runnable jobs"

Verify golden job items are hidden in resultset.

rbillings:
This was merged.

PR #69, opened 2016-05-18
Add new jobs

rbillings:
Addresses issue #67 , test for adding new jobs.

### DIFF
--- a/tests/test_add_new_jobs.py
+++ b/tests/test_add_new_jobs.py
@@ -1,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.treeherder import TreeherderPage
+
+
+@pytest.mark.nondestructive
+def test_add_new_jobs(base_url, selenium):
+    page = TreeherderPage(base_url, selenium).open()
+    assert len(page.result_sets) >= 1
+    page.result_sets[0].view()
+    page.result_sets[0].add_new_jobs()
+    assert len(page.result_sets[0].runnable_jobs) > 0
+
+
+@pytest.mark.nondestructive
+def test_hide_runnable_jobs(base_url, selenium):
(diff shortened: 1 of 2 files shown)


tests/test_add_new_jobs.py
```diff
@@ -0,0 +1,26 @@
...
+    page.result_sets[0].add_new_jobs()
+    assert len(page.result_sets[0].runnable_jobs) > 0
+
+@pytest.mark.nondestructive
+def test_hide_runnable_jobs(base_url, selenium):
+    page = TreeherderPage(base_url, selenium).open()
+    assert len(page.result_sets) > 1
+    page.result_sets[0].view()
+    page.result_sets[0].add_new_jobs()
+    assert 0 < len(page.result_sets[0].runnable_jobs)
```
davehunt: Nit: We should keep this consistent with the assert from `test_add_new_jobs`, and assert `> 0`

tests/test_add_new_jobs.py
```diff
@@ -0,0 +1,26 @@
...
+
+import pytest
+
+from pages.treeherder import TreeherderPage
+
+
+@pytest.mark.nondestructive
+def test_add_new_jobs(base_url, selenium):
+    page = TreeherderPage(base_url, selenium).open()
+    assert len(page.result_sets) > 1
```
davehunt: Nit: Both here and on line 21, I think we can assert that there's at least one result set, as we only use the first one. You could make this `> 0` or `>= 1`.

tests/test_add_new_jobs.py
```diff
@@ -0,0 +1,26 @@
...
+    page.result_sets[0].view()
+    page.result_sets[0].add_new_jobs()
+    assert len(page.result_sets[0].runnable_jobs) > 0
+
```
davehunt: We need an additional blank line between the two tests.

davehunt:
Looks good, just a couple of nits that aren't blockers. Also there's a flake8 issue. Consider this r+ with at least the flake8 issue addressed. Note that we should squash the commits on merge if someone else takes care of this.

rbillings:
Squashed after fixing flake8 issues.

Issue #62, opened 2016-05-04
Test: Filter platform and jobs results


rbillings:
Create a test that filters results based on platform or jobs.

1) Open Treeherder page
2) In the "Filter platforms and jobs" search box, add data for a platform or job name
3) Submit search by clicking enter

Verify that jobs are updated correctly, reflecting the search query.

4) Click the "x" at the end of the saved search query

Verify that search parameters are cleared and all jobs display

PR #70, opened 2016-05-24
Test to filter platform

rbillings:
Addresses Issue #62

### DIFF
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -69,14 +69,14 @@ class TreeherderPage(Base):
         self.wait.until(lambda s: len(self.result_sets) == 60)
         return self
 
-    def filter_clear(self):
+    def clear_filter(self):
         self.selenium.find_element(*self._clear_filter_locator).click()
 
-    def filter_search(self, term):
+    def filter_by(self, term):
         el = self.selenium.find_element(*self._quick_filter_locator)
         el.send_keys(term)
         el.send_keys(Keys.RETURN)
-        self.wait_for_page_to_load()
+        self.wait.until(lambda s: self.result_sets)
 
     def open_next_unclassified_failure(self):
         el = self.find_element(*self._result_sets_locator)
@@ -123,9 +123,14 @@ class TreeherderPage(Base):
         _hide_runnable_jobs_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(2) > a')
         _jobs_locator = (By.CLASS_NAME, 'job-btn')
         _pin_all_jobs_locator = (By.CLASS_NAME, 'pin-all-jobs-btn')
+        _platform_locator = (By.CLASS_NAME, 'platform')
         _runnable_jobs_locator = (By.CSS_SELECTOR, '.runnable-job-btn.filter-shown')
         _set_bottom_of_range_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(8) > a')
         _set_top_of_range_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(7) > a')
+
+        @property
+        def builds(self):
+            return [self.Build(self.page, root=el) for el in self.find_elements(*self._platform_locator)]
 
         @property
         def datestamp(self):
@@ -170,6 +175,13 @@ class TreeherderPage(Base):
 
         def view(self):
             return self.find_element(*self._datestamp_locator).click()
+
+        class Build(Region):
+            _platform_name_locator = (By.CSS_SELECTOR, 'td:nth-child(1) > span:nth-child(1)')
+
+            @property
+            def platform_name(self):
+                return self.find_element(*self._platform_name_locator).text
 
         class Job(Region):
 
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -15,10 +15,12 @@ from pages.base import Base
 class TreeherderPage(Base):
 
     _active_watched_repo_locator = (By.CSS_SELECTOR, '#watched-repo-navbar button.active')
+    _clear_filter_locator = (By.ID, 'quick-filter-clear-button')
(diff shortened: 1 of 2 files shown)


rbillings:
I'm wondering if this needs additional assertions confirming the presence of the search term in the displayed jobs.

pages/treeherder.py
```diff
@@ -52,6 +54,15 @@ def unchecked_repos(self):
     def unclassified_failure_count(self):
         return int(self.selenium.find_element(*self._unclassified_failure_count_locator).text)
 
+    def filter_clear(self):
```
davehunt: Nit: I'd call this `clear_filter`

pages/treeherder.py
```diff
@@ -52,6 +54,15 @@ def unchecked_repos(self):
...
+        self.selenium.find_element(*self._clear_filter_locator).click()
+
+    def filter_search(self, term):
```
davehunt: Nit: I'd call this `filter_by`

tests/test_filter_jobs.py
```diff
@@ -0,0 +1,20 @@
...
+from pages.treeherder import TreeherderPage
+
+
+@pytest.mark.nondestructive
+def test_filter_jobs(base_url, selenium):
+    """Open resultset page and filter for platform"""
+    term = u'Linux'
+    page = TreeherderPage(base_url, selenium).open()
+    page.filter_search(term)
+    assert "&filter-searchStr=Linux" in page.selenium.current_url
```
davehunt: I think a more valuable check would be that there are only Linux jobs listed. This would probably require a check that there are not just linux jobs before we filter (which is going to be dependent on data). Something like:

```python
platform = 'linux'
initial_builds = page.result_sets[0].builds
initial_platforms = [b.platform for b in initial_builds]

page.filter_by(platform)
filtered_builds = page.result_sets[0].builds
filtered_platforms = [b.platform for b in filtered_builds]
assert len(initial_builds) > len(filtered_builds)
assert all([platform.lower() in p.lower() for p in filtered_platforms])

page.clear_filter()
unfiltered_builds = page.result_sets[0].builds
unfiltered_platforms = [b.platform for b in unfiltered_builds]
assert len(initial_builds) == len(unfiltered_builds)
assert initial_platforms == unfiltered_platforms
```

This would require a `Builds` region under `ResultSet` with a `platform` property. We probably should then move `jobs` to beneath `Builds`, though we could keep a `jobs` property under `ResultSet` that returned all the jobs for all the builds.

rbillings: I've been working on this, and unfortunately there is no query based on platform that doesn't return multiple pages of results. I'm still tinkering with it, I just wanted to comment to show this hasn't been totally lost.

rbillings:
This is ready for review, @mozilla/web-qa-sorcerers

pages/treeherder.py
```diff
@@ -49,6 +51,15 @@ def unchecked_repos(self):
     def unclassified_failure_count(self):
         return int(self.find_element(*self._unclassified_failure_count_locator).text)
 
+    def clear_filter(self):
```
m8ttyB: For readability would `click_clear_filter` be appropriate?

rbillings: Actually I just realized that I no longer use this with the updated verification. For my 2 cents I wouldn't change it to add "click", but that's me.

tests/test_filter_jobs.py
```diff
@@ -0,0 +1,21 @@
...
+    """Open resultset page and filter for platform"""
+    page = TreeherderPage(selenium, base_url).open()
+    platform = u'Linux'
+
+    page.filter_by(platform)
+    assert platform in page.result_sets[0].builds[0].platform_name
+
+    platform2 = u'Windows'
+    page.filter_by(platform2)
+    assert platform not in page.result_sets[0].builds[0].platform_name
```
m8ttyB: Should you also do an assert for the 2nd platform?
`assert platform2 in page.result_sets[0].builds[0].platform_name`

pages/treeherder.py
m8ttyB: 👍 Should you remove the method if it's not in use?

tests/test_filter_jobs.py
rbillings: Verification on this one is tricky since the default page will have all of the potential search options, but too many to list as an include. I didn't see the value of making a second assert that confirms a positive display- as opposed to one positive & one negative, but I'm open to the idea.

m8ttyB: It depends on how much Javascript magic is used to apply and remove filters. Sometimes a boundary issue can be found by asserting the correct filter applied.

rbillings: Clarification: It isn't easy to make a confirmation of the state of the filter before searching as there are too many options to confirm each one. So I went with one positive and one negative assert to cover different scenarios.

rbillings:
Rebased with master, ready for review, @mozilla/web-qa-sorcerers ?

pages/treeherder.py
davehunt: +1 on removing this if it's not used
-1 on renaming to add 'click' if we decide to use/keep it

pages/treeherder.py
```diff
@@ -52,6 +54,15 @@ def unchecked_repos(self):
...
+        self.selenium.find_element(*self._clear_filter_locator).click()
+
+    def filter_by(self, term):
+        el = self.selenium.find_element(*self._quick_filter_locator)
+        el.send_keys(term)
+        el.send_keys(Keys.RETURN)
+        self.wait.until(lambda s: self.result_sets)
```
davehunt: Wouldn't this wait condition also be evaluated as true before we filter?

pages/treeherder.py
davehunt: I see that it is used :)

davehunt:
r+

PR #71, opened 2016-05-26
Migrate to PyPOM

davehunt:
@rbillings r?


pages/treeherder.py
```diff
@@ -148,25 +141,23 @@ class JobDetails(PageRegion):
...
-            return self.selenium.find_element(*self._job_result_status_locator).text
+            return self.find_element(*self._job_result_status_locator).text
 
         def open_logviewer(self):
-            self.selenium.find_element(*self._job_details_panel_locator).send_keys('l')
+            self.find_element(*self._job_details_panel_locator).send_keys('l')
             window_handles = self.selenium.window_handles
             for handle in window_handles:
                 self.selenium.switch_to.window(handle)
-                self.selenium.implicitly_wait(1)
```
davehunt: This change is unrelated to PyPOM, but we shouldn't set an implicit wait anywhere in this suite, due to our reliance on explicit waits. In this particular case, it's not actually doing anything here, but could have an impact on any code that runs after this.

rbillings:
Changes look good, tests pass, r+

PR #80, opened 2016-08-31
Changing to upper case. Otherwise will not work

MichalTHEDUDE:


rbillings:
Great catch! My only comment is for future pull requests: please create a branch for your pull request so it's not merging from the master branch.

PR #81, opened 2016-09-02
Test the Filter Panel

MichalTHEDUDE:
All test have passed. 
IMO code is unoptimized. All feedback appreciated :)

### DIFF
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import itertools
 import random
 
 from pypom import Page, Region
@@ -16,6 +17,13 @@ class TreeherderPage(Base):
 
     _active_watched_repo_locator = (By.CSS_SELECTOR, '#watched-repo-navbar button.active')
     _clear_filter_locator = (By.ID, 'quick-filter-clear-button')
+    _close_the_job_panel_locator = (By.CSS_SELECTOR, '.info-panel-navbar-controls > li:nth-child(2)')
+    _filter_panel_locator = (By.CSS_SELECTOR, 'span.navbar-right > span:nth-child(4)')
+    _filter_panel_all_failures_locator = (By.CSS_SELECTOR, '.pull-right input')
+    _filter_panel_testfailed_failures_locator = (By.ID, 'testfailed')
+    _filter_panel_busted_failures_locator = (By.ID, 'busted')
+    _filter_panel_exception_failures_locator = (By.ID, 'exception')
+    _filter_panel_reset_locator = (By.CSS_SELECTOR, '.pull-right span:nth-child(3)')
     _mozilla_central_repo_locator = (By.CSS_SELECTOR, '#th-global-navbar-top a[href*="mozilla-central"]')
     _next_ten_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(1)')
     _next_twenty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(2)')
@@ -26,6 +34,7 @@ class TreeherderPage(Base):
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
     _unchecked_repos_links_locator = (By.CSS_SELECTOR, '#repoLabel + .dropdown-menu .dropdown-checkbox:not([checked]) + .dropdown-link')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
+    _unclassified_failure_filter_locator = (By.CSS_SELECTOR, '.btn-unclassified-failures')
 
     def wait_for_page_to_load(self):
         self.wait.until(lambda s: self.unclassified_failure_count > 0)
@@ -34,6 +43,10 @@ class TreeherderPage(Base):
     @property
     def active_watched_repo(self):
         return self.find_element(*self._active_watched_repo_locator).text
+
+    @property
+    def all_jobs(self):
+        return list(itertools.chain.from_iterable([r.jobs for r in self.result_sets]))
 
     @property
     def job_details(self):
@@ -83,6 +96,9 @@ class TreeherderPage(Base):
         el.send_keys(Keys.RETURN)
         self.wait.until(lambda s: self.result_sets)
 
+    def filter_unclassified_jobs(self):
+        self.find_element(*self._unclassified_failure_filter_locator).click()
+
     def open_next_unclassified_failure(self):
         el = self.find_element(*self._result_sets_locator)
         self.wait.until(EC.visibility_of(el))
@@ -118,6 +134,53 @@ class TreeherderPage(Base):
         self.wait.until(lambda s: self._active_watched_repo_locator == repo_name)
         return repo_name
 
+    def click_on_filters_panel(self):
+        self.find_element(*self._filter_panel_locator).click()
+
+    def close_the_job_panel(self):
+        self.find_element(*self._close_the_job_panel_locator).click()
+
+    def deselect_all_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_all_failures_locator).click()
+
+    def reset_filters(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_reset_locator).click()
+
+    def select_testfailed_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_testfailed_failures_locator).click()
+
+    def deselect_testfailed_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_testfailed_failures_locator).click()
+
+    def select_busted_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_busted_failures_locator).click()
+
+    def deselect_busted_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_busted_failures_locator).click()
+
+    def select_exception_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_exception_failures_locator).click()
+
+    def deselect_exception_failures(self):
+        """Filters Panel must be opened"""
+        self.find_element(*self._filter_panel_exception_failures_locator).click()
+
+    def checkbox_testfailed_is_selected(self):
+        return self.find_element(*self._filter_panel_testfailed_failures_locator).is_selected()
+
+    def checkbox_busted_is_selected(self):
+        return self.find_element(*self._filter_panel_busted_failures_locator).is_selected()
+
+    def checkbox_exception_is_selected(self):
+        return self.find_element(*self._filter_panel_exception_failures_locator).is_selected()
+
     class ResultSet(Region):
 
         _add_new_job_locator = (By.CSS_SELECTOR, '.open ul > li a')
@@ -126,7 +189,7 @@ class TreeherderPage(Base):
         _expanded_group_content_locator = (By.CSS_SELECTOR, '.group-job-list[style="display: inline;"]')
         _group_content_locator = (By.CSS_SELECTOR, 'span.group-count-list .btn')
         _hide_runnable_jobs_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(2) > a')
-        _jobs_locator = (By.CLASS_NAME, 'job-btn')
+        _jobs_locator = (By.CSS_SELECTOR, '.job-btn.filter-shown')
         _pin_all_jobs_locator = (By.CLASS_NAME, 'pin-all-jobs-btn')
         _platform_locator = (By.CLASS_NAME, 'platform')
         _runnable_jobs_locator = (By.CSS_SELECTOR, '.runnable-job-btn.filter-shown')
--- a/tests/test_filter_panel.py
+++ b/tests/test_filter_panel.py
@@ -1,0 +1,63 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.treeherder import TreeherderPage
+
+
+@pytest.mark.nondestructive
+def test_filter_by_test_status(base_url, selenium):
+    """Open Treeherder page, open Filters Panel, select one filter,
+    verify results"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.filter_unclassified_jobs()
+    page.click_on_filters_panel()
+
+    # Test 'testfailed' unclassified failures
+    page.deselect_busted_failures()
+    page.deselect_exception_failures()
+    if len(page.all_jobs) > 0:
+        page.open_next_unclassified_failure()
+        assert 'testfailed' == page.job_details.job_result_status
+
+    # Test 'busted' unclassified failures
+    page.select_busted_failures()
+    page.deselect_testfailed_failures()
+    if len(page.all_jobs) > 0:
+        page.close_the_job_panel()
+        page.open_next_unclassified_failure()
+        assert 'busted' == page.job_details.job_result_status
+
+    # Test 'exception' unclassified failures
+    page.select_exception_failures()
+    page.deselect_busted_failures()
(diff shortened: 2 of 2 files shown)


tests/test_filter_panels.py
```diff
@@ -0,0 +1,91 @@
...
+    # Get next 50 jobs to see 'exception' unclassified failures
+    page.get_next_fifty_results()
+    page.deselect_all_failures()
+
+    # Check 'testfailed' unclassified failures
+    page.select_testfailed_failures()
+    page.open_next_unclassified_failure()
+    teststatus = page.job_details.job_result_status
+    assert teststatus in ['testfailed']
+    page.select_testfailed_failures()
```
rbillings: This seems like an unnecessary or duplicate step after you've made the assert, and could be removed.

rbillings: Oh I see - it's de-selecting the testfailed checkbox.  Adding to the comment on line 21 that you'll remove the check would clarify this.

tests/test_filter_panels.py
```diff
@@ -0,0 +1,91 @@
...
+    are selected"""
+    page = TreeherderPage(selenium, base_url).open()
+    assert page.unclassified_failure_count > 0
+
+    page.open_next_unclassified_failure()
+    teststatus = page.job_details.job_result_status
+    assert teststatus in ['busted', 'testfailed', 'exception']
+
+    page.deselect_all_failures()
+    with pytest.raises(Exception) as e_info:   # Waiting for TimeoutException
```
rbillings: It seems like there should be a step after deselecting all failures where you try to find the next unclassified failure- thus prompting the check.

tests/test_filter_panels.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.nondestructive
+def test_reset_filter_panels2(base_url, selenium):
+    """Version 2
+
+    Open Treeherder page, search for next unclassified failure,
+    open Filters Panel, disable all failures, check that there are no failures
+    to select, click reset button and verify that default checkboxes
+    are selected"""
```
rbillings: I don't understand the purpose of having a second reset filters test.

rbillings:
This is a really excellent start @MichalTHEDUDE ! I made a few comments, but the foundation of what you've written is very well done. It would be great if @davehunt would also take a look in regards to naming conventions, etc.

tests/test_filter_panels.py
MichalTHEDUDE: Changed 'Check' it to 'Test".

tests/test_filter_panels.py
MichalTHEDUDE: You are right! But will remove Version1 and go with Version2.

tests/test_filter_panels.py
MichalTHEDUDE: I have hesitated a little bit which `pytest.raise` version to choose. But will proceed with Version2.

MichalTHEDUDE:
Thanks @rbillings for your feedback! It is really great to hear that :).

Have question for `test_edit_filter_panels`. This test take a long time to run. Is there any better way to write this test, to optimize it?
Would it be a good idea to write a class for FilterPanel? Like for JobDetails or Pinboard?

rbillings:
The changes look good. I'm not sure how to optimize this test- perhaps @davehunt will have an idea on that. 

As for your suggestion of a class, we're following the PyPOM region model which requires repeating elements so the filter would not qualify. You can read more about it here: http://pypom.readthedocs.io/en/latest/user_guide.html#regions

Lastly you will want to rebase to fix up merge conflicts.

rbillings:
The changes look good. I'm not sure how to optimize this test- perhaps @davehunt will have an idea on that. 

As for your suggestion of a class, we're following the PyPOM region model which requires repeating elements so the filter would not qualify. You can read more about it here: http://pypom.readthedocs.io/en/latest/user_guide.html#regions

Lastly you will want to rebase to fix up merge conflicts.

rbillings:
The changes look good. I'm not sure how to optimize this test- perhaps @davehunt will have an idea on that. 

As for your suggestion of a class, we're following the PyPOM region model which requires repeating elements so the filter would not qualify. You can read more about it here: http://pypom.readthedocs.io/en/latest/user_guide.html#regions

Lastly you will want to rebase to fix up merge conflicts.

rbillings:
The changes look good. I'm not sure how to optimize this test- perhaps @davehunt will have an idea on that. 

As for your suggestion of a class, we're following the PyPOM region model which requires repeating elements so the filter would not qualify. You can read more about it here: http://pypom.readthedocs.io/en/latest/user_guide.html#regions

Lastly you will want to rebase to fix up merge conflicts.

rbillings:
The changes look good. I'm not sure how to optimize this test- perhaps @davehunt will have an idea on that. 

As for your suggestion of a class, we're following the PyPOM region model which requires repeating elements so the filter would not qualify. You can read more about it here: http://pypom.readthedocs.io/en/latest/user_guide.html#regions

Lastly you will want to rebase to fix up merge conflicts.

rbillings:
ugh GitHub! Sorry for repeating comments

MichalTHEDUDE:
No worries.
I will rebase some commits and push the changes.

Thanks!

MichalTHEDUDE:
I rebased the commits, but I've messed something up.... Because During my work I pull'ed your `test_get_next_results.py` changes and now my PR shows that I've implemented this.

Is there a way to undo those?

davehunt:
> Is there a way to undo those?

You might be able to use `git reflog` to revert to before your rebase and try again. See [this answer](http://stackoverflow.com/questions/134882/undoing-a-git-rebase#135614) on StackOverflow for some details. I would advise saving your changes somewhere first though, in case you're unable to revert and lose them.

tests/test_filter_panels.py
```diff
@@ -0,0 +1,63 @@
...
+
+@pytest.mark.nondestructive
+def test_edit_filter_panels(base_url, selenium):
+    """Open Treeherder page, open Filters Panel, select one filter,
+    verify results"""
+    page = TreeherderPage(selenium, base_url).open()
+    assert page.unclassified_failure_count > 0
+
+    # Get next 50 jobs to see 'exception' unclassified failures
+    page.get_next_fifty_results()
```
davehunt: Why do we need to load the next 50 jobs?

tests/test_filter_panels.py
```diff
@@ -0,0 +1,63 @@
...
+    page.deselect_all_failures()
+
+    # Test 'testfailed' unclassified failures
+    page.select_testfailed_failures()
+    page.open_next_unclassified_failure()
+    teststatus = page.job_details.job_result_status
+    assert teststatus in ['testfailed']
```
davehunt: Can we change this to `assert 'testfailed' == page.job_details.job_result_status`?

tests/test_filter_panels.py
davehunt: It would read better if we had separate methods for selecting/deselecting by failure status. This also give us the benefit of checking the initial state. We can also manage this in a page region for the filters, to avoid overloading the page object. For example the test might look like:

```python
filters = page.open_filters()
filters.failures.hide_all()  # hide all failures
filters.failures.show_testfailed()  # show testfailed failures
page.open_next_unclassified_failure()
assert 'testfailed' == page.job_details.job_result_status
filters.failures.hide_testfailed()
...
```

tests/test_filter_panels.py
```diff
@@ -0,0 +1,63 @@
...
+    page.select_testfailed_failures()
+
+    # Test 'busted' unclassified failures
+    page.select_busted_failures()
+    page.open_next_unclassified_failure()
+    teststatus = page.job_details.job_result_status
+    assert teststatus in ['busted']
+    page.select_busted_failures()
```
davehunt: I think this should be a separate test. As you point out the test takes a long time to run, so splitting it into a test for each status will allow them to run in parallel. I do wonder if we need to test all of the status values though, as the functionality should be identical. If we do decide to cover multiple status values it could even make sense to make this a parameterised test to reduce duplication.

tests/test_filter_panels.py
```diff
@@ -0,0 +1,63 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.treeherder import TreeherderPage
+
+
+@pytest.mark.nondestructive
+def test_edit_filter_panels(base_url, selenium):
```
davehunt: I'm not sure this is the most representative test name. Perhaps `test_filter_by_test_status` would be more appropriate?

tests/test_filter_panels.py
```diff
@@ -0,0 +1,63 @@
...
+    page = TreeherderPage(selenium, base_url).open()
+    assert page.unclassified_failure_count > 0
+
+    page.open_next_unclassified_failure()
+    teststatus = page.job_details.job_result_status
+    assert teststatus in ['busted', 'testfailed', 'exception']
+
+    page.deselect_all_failures()
+    with pytest.raises(Exception):  # Waiting for TimeoutException
+        page.open_next_unclassified_failure()
```
davehunt: We should avoid the cost of a timeout if possible. If this test is about testing the filter reset reset then we shouldn't need to assert that we're unable to open an unclassified failure as that's a specific check for filtering out failures, and could be a test it it's own right. I would consider capturing the total number of jobs before modifying the filters, ensuring it is less after the filter is applied, and finally checking that it's back to the original number after the reset.

tests/test_filter_panels.py
```diff
@@ -0,0 +1,63 @@
...
+
+    page.reset_filters()
+    assert page.checkbox_testfailed_is_selected()
+    assert page.checkbox_busted_is_selected()
+    assert page.checkbox_exception_is_selected()
```
davehunt: These checks are okay, but we should at least check they are also false somewhere to confirm they're checking what they should be.

tests/test_filter_panels.py
MichalTHEDUDE: IMO yes. Because it is hard to catch 'exception' failures (as they do not appear in first 10 jobs).

tests/test_filter_panels.py
MichalTHEDUDE: It is up to you guys. Maybe @rbillings could help us?

tests/test_filter_panels.py
davehunt: We should avoid having tests that are so heavily reliant on specific data they can't control. I would suggest avoiding this check if exceptions are so uncommon. If there happened to be none in the last 60 jobs then this test would fail even if there is no genuine regression.

MichalTHEDUDE: True. So I will delete this test. Should I split rest of the tests cases (testfailed, busted) for different functions?

davehunt: I would argue that testing one status filter is enough to cover the functionality, but let's see what @rbillings thinks.

MichalTHEDUDE:
Fixed issue with Git! But now it seems that all comments are collapsed :(.
Sorry Guys!

rbillings:
There are a number of comments, so I’ll try to address them here:
- I agree there is no need to get next 50 jobs as there is no guarantee that an exception will display- even though it does increase the likelihood of it happening.
- Regarding the question of a single vs. multiple tests: It would run faster as multiple, but the functionality is so similar. I agree that one status filter is enough to cover the functionality. Ideally I would check the original page to see if there is a job of the correct status [each one] and if not, then skip to the next check. That way there would be no false failures, and have it within a single test.
- I do not agree that counting the filters and the counting again after the filter is applied is a good test. With test failed, for example, the count will be the same. I am working on a test that verifies the status of jobs, as opposed to the count of a job. This will work as an alternate assuming you want to wait for me to complete it. [I will focus on it tomorrow]
- I think it is a good idea to have separate methods to select/deselect checkboxes for readability. I didn’t think about doing a page region in the way Dave suggested, but I like the idea of a page region like that.

rbillings:
You can check out my pull request that includes finding and asserting the presence of 'test failed' status jobs here: https://github.com/mozilla/treeherder-tests/pull/83

davehunt:
> now it seems that all comments are collapsed

They're still there, and can be expanded and replied to.

davehunt:
> I do not agree that counting the filters and the counting again after the filter is applied is a good test. With test failed, for example, the count will be the same. I am working on a test that verifies the status of jobs, as opposed to the count of a job. This will work as an alternate assuming you want to wait for me to complete it.

Perhaps my suggestion was unclear. I'll see if I can provide a demonstration instead.

- Load https://treeherder.allizom.org
- document.querySelectorAll('.job-btn.filter-shown').length = 1481 (all displayed jobs)
- document.querySelectorAll('.job-btn.btn-orange.filter-shown').length = 133 (displayed testfailed jobs)
- Click Filters > failures > all (deselect)
- document.querySelectorAll('.job-btn.filter-shown').length = 1317 (less jobs shown)
- document.querySelectorAll('.job-btn.btn-orange.filter-shown').length = 0 (none displayed)
- Click Filters > failures > testfailed (select)
- document.querySelectorAll('.job-btn.filter-shown').length = 1473 (more jobs shown, but not exceeding original count)
- document.querySelectorAll('.job-btn.btn-orange.filter-shown').length = 133 (they're all back)
- document.querySelectorAll('.job-btn.btn-orange.filter-shown')[0].getAttribute('title') = "desktop-test-linux64/opt-reftest-no-accel-e10s-3 - testfailed (15 mins)" (first shown unclassified failure has 'testfailed' in title - we could check them all, or we could pick one at random, or just check the first one).

Note that more content is loaded dynamically while the page is open, so I'm not sure we can make these tests too dependent on actual numbers. Ideally we would test this functionality with a dataset we control - has there been any movement on that approach @rbillings?

rbillings:
Next steps:
- remove adding 50 jobs
- create a single test, it’ll be slower but the functionality is so similar it should all be in one test
- add methods to select or deselect checkboxes 
- on initial page load you can find the list of all_jobs and then filter and make sure they are not equal. This code is in the PR link below*
- you can use davehunt’s suggestions on verifying the title of the jobs as the job status is in that.

* https://github.com/mozilla/treeherder-tests/pull/83/files#diff-530cb5e37852eb76641f23e0e45f6ccdL35

MichalTHEDUDE:
Guys thanks for your feedback and hints :).
I've made two versions according to your suggestions, but I couldn't test it as #83 is still in PR (and I've used code from this issue).

Still don't know what to do with 'exception' failures (that's why I've added `if` in Version 3).

What do you think of those versions? Does them make sense?

rbillings:
Good improvements! You can add the code from #83 and test it locally - I did and the tests pass. It doesn't matter who adds the locator first. 

I think Version 3 is great - I would actually add the "if" statement to the "busted" category also as the only type that is virtually guaranteed to show is test failed. I like that as a workaround. This is close to being ready to merge!

MichalTHEDUDE:
I've added the 'if' statement to the 'busted' failures.

Should @davehunt also take a look at the code?

rbillings:
The changes with the If statements look good. However, I just ran the test and it is failing for me because it's not clicking the deselect all failures checkbox. Does this pass for you locally? Also:

- You will need to add the code from the other PR or these tests will never pass.
- You have some flake8 errors that need fixing
- You'll need to rebase with master

MichalTHEDUDE:
Just added code to run all the tests. 

But still need to work on `test_filter_by_test_status`, as there are problems with `exception` failures (TimeoutException - no items to display). 

Now I'm seeing some conflicts :(. I will rebase with master and then we will see.

MichalTHEDUDE:
- I'm not happy using `time.sleep(1)` but for now it is the only option for `job_results_status` to load. Otherwise it will take previous job status and return assertion error.
- I have rebase to master (3 commits behind current master).
- How can I check where are conflicts? As Travis CI suggesting.

rbillings:
Hi!
- I agree about not using time.sleep. I tested this a number of ways and see that even with adding more waits for the job details panel it's picking up previous results. An idea on how to resolve this is to close the panel in between tests, creating a "close_logviewer" method.
- Checking conflicts is easy in github. Trying "git status" and that should show you the conflicts locally. Make sure your master branch is up to date first.

MichalTHEDUDE:
- added `close job panel` function and implemented into the tests.
- all tests are passing now (without `time.sleep()`).
- I've rebased with master, so no conflict should appear.

rbillings:
This looks good!

MichalTHEDUDE:
No way it is merged! Aww yeah! Thanks! :D

rbillings:
Thanks for your hard work & persistence, @MichalTHEDUDE !!

Issue #49, opened 2016-04-27
Test: View only unclassified jobs


rbillings:
Create a test which changes the display to only show unclassified jobs.

1) Open treeherder
2) Click on the "(n) unclassified" button in the top nav

Verify that only unclassified failures are visible in the job table.

PR #83, opened 2016-09-08
Add test to verify testfailed status jobs in unclassified filter

rbillings:
This addresses issue #49 . I have chosen to only verify "testfailed" status jobs as it is much more common, instead of including also the 'busted' and 'test exception' jobs.


pages/treeherder.py
```diff
@@ -115,6 +119,9 @@ class ResultSet(Region):
         _runnable_jobs_locator = (By.CSS_SELECTOR, '.runnable-job-btn.filter-shown')
         _set_bottom_of_range_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(8) > a')
         _set_top_of_range_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(7) > a')
+        _test_busted_job_title_locator = (By.CSS_SELECTOR, 'span.group-job-list .btn-purple')
```
MichalTHEDUDE: I'm just wondering if we should use this CSS selector `span.group-job-list .btn-purple`. 
Because sometimes if you have no name like 'Mochitest', 'TestCluster', 'Reftest' in front of Failure color button there is `td` tag, instead of `span`.

Plus `.btn-purple` is for `exception` failures :)

pages/treeherder.py
```diff
@@ -115,6 +119,9 @@ class ResultSet(Region):
...
+        _test_exception_job_title_locator = (By.CSS_SELECTOR, 'span.group-job-list .btn-red')
```
MichalTHEDUDE: `.btn-red` is for `busted` failures.

pages/treeherder.py
MichalTHEDUDE: Maybe it will be better to have CSS_SELECTOR like `button.btn-red`.
What do you think?

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,13 @@ def test_open_unclassified_failure_log(base_url, selenium):
     treeherder_page.open_next_unclassified_failure()
     logviewer_page = treeherder_page.job_details.open_logviewer()
     assert logviewer_page.is_job_status_visible
+
+
+@pytest.mark.nondestructive
+def test_view_unclassified_jobs(base_url, selenium):
+    page = TreeherderPage(selenium, base_url).open()
+    assert len(page.result_sets[0].testfailed_job_title) >= 0
```
davehunt: This appears to be checking that the title for the first testfailed job for the first result set is non-zero. The test name suggests that we're testing the unclassified jobs filter, and I'm not sure this assertion is appropriate. Wouldn't this fail if the first result set didn't have any test failures? Also, we already have a jobs property - could we use that to return the title, and filter it based on visibility or status?

pages/treeherder.py
davehunt: This isn't used.

pages/treeherder.py
davehunt: This isn't used.

pages/treeherder.py
rbillings: Originally I had locators for each type of unclassified failure. I really meant to delete them and to verify only 'test failed' ones.

tests/test_unclassified_job.py
rbillings: The test is to check unclassified failures, and yes it will fail if there aren't any test failures. As with a number of other Treeherder tests I'm not entirely sure how to gracefully fail if the data I'm looking for doesn't happen to exist. 
I'm unclear what you mean about using the jobs property in this way. The filter button returns all types of unclassified failures. I guess to be thorough I should return to verifying that all types of unclassified failures display after the filter, but again there's the test data problem.

davehunt: > yes it will fail if there aren't any test failures

You could perhaps not limit the test to just the first result set, which would increase the chances of there being testfailed jobs.

> I'm unclear what you mean about using the jobs property in this way

I mean something like `page.result_sets[0].jobs[0].title` rather than creating a new property for a certain job status on the page object. If you used this you should even be able to iterate or compile a list of all jobs displayed and ensure that they are unclassified. From your asserts I gather than a job with a status of testfailed means the same as unclassified? Otherwise I'm not sure how we're checking the filter is working as expected.

rbillings:
Added a method to find the title for jobs. It works when under the ResultSet region, but not when under the Job region - I need to investigate more.

rbillings:
Ready for review, @mozilla/web-qa-sorcerers

stephendonner:
Although I'm unable to give a proper review (just not technically capable at this level, yet), I've run this several times, and keep hitting the same failure:

http://pastebin.com/UgnN1Mwv

Perhaps telling, though, is that I'm sporadically seeing other tests now fail, so maybe there's a more-general issue right now with Treeherder staging?

http://pastebin.com/6wR6BCy1

pages/treeherder.py
```diff
@@ -78,6 +84,11 @@ def filter_by(self, term):
         el.send_keys(Keys.RETURN)
         self.wait.until(lambda s: self.result_sets)
 
+    def filter_unclassified_jobs(self):
+        self.find_element(*self._unclassified_failure_filter_locator).click()
+        self.wait.until(lambda s: len(self.result_sets) == 10)
```
davehunt: This wait is not specific to the action of filtering unclassified jobs. If you first click on show next 50 results, then filter unclassified jobs, then you'll have more than 10 result sets.

pages/treeherder.py
```diff
@@ -78,6 +84,11 @@ def filter_by(self, term):
...
+        return self
```
davehunt: No need to return the page object here.

pages/treeherder.py
```diff
@@ -117,11 +128,12 @@ class ResultSet(Region):
 
         _add_new_job_locator = (By.CSS_SELECTOR, '.open ul > li a')
         _datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span a')
+        _displayed_job_title_locator = (By.CSS_SELECTOR, '.job-btn.filter-shown')
         _dropdown_toggle_locator = (By.CLASS_NAME, 'dropdown-toggle')
         _expanded_group_content_locator = (By.CSS_SELECTOR, '.group-job-list[style="display: inline;"]')
         _group_content_locator = (By.CSS_SELECTOR, 'span.group-count-list .btn')
         _hide_runnable_jobs_locator = (By.CSS_SELECTOR, '.open ul > li:nth-child(2) > a')
-        _jobs_locator = (By.CLASS_NAME, 'job-btn')
+        _jobs_locator = (By.CLASS_NAME, 'job-btn.filter-shown')
```
davehunt: This should use the `CSS_SELECTOR` strategy and value `.job-btn.filter-shown`

pages/treeherder.py
```diff
@@ -117,11 +128,12 @@ class ResultSet(Region):
 
         _add_new_job_locator = (By.CSS_SELECTOR, '.open ul > li a')
         _datestamp_locator = (By.CSS_SELECTOR, '.result-set-title-left > span a')
+        _displayed_job_title_locator = (By.CSS_SELECTOR, '.job-btn.filter-shown')
```
davehunt: Duplicate of `_jobs_locator` and should be removed.

pages/treeherder.py
```diff
@@ -137,6 +149,10 @@ def datestamp(self):
             return self.find_element(*self._datestamp_locator).text
 
         @property
+        def displayed_job_title(self):
+            return self.find_element(*self._displayed_job_title_locator).get_attribute('title')
```
davehunt: This should be removed in favour of `result_set[0].jobs[0].title`

pages/treeherder.py
```diff
@@ -185,14 +201,18 @@ def platform_name(self):
 
         class Job(Region):
 
-            def click(self):
-                self.root.click()
-                self.wait.until(lambda _: self.page.job_details.job_result_status)
```
davehunt: Was this removed intentionally?

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,23 @@ def test_open_unclassified_failure_log(base_url, selenium):
     treeherder_page.open_next_unclassified_failure()
     logviewer_page = treeherder_page.job_details.open_logviewer()
     assert logviewer_page.is_job_status_visible
+
+
+@pytest.mark.nondestructive
+def test_view_unclassified_jobs(base_url, selenium):
+    page = TreeherderPage(selenium, base_url).open()
+    all_jobs = list(page.all_jobs)
```
davehunt: `page.all_jobs` already returns a list, so we can remove `list()`

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,23 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+
+    page.filter_unclassified_jobs()
+    filtered_jobs = list(page.all_jobs)
```
davehunt: As above, we can remove `list()`

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,23 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+
+    assert not all_jobs == filtered_jobs
+    if 'testfailed' in jobstatus:
+        assert 'testfailed' in jobstatus
+
+    if 'busted' in jobstatus:
+        assert 'busted' in jobstatus
+
+    if 'exception' in jobstatus:
+        assert 'exception' in jobstatus
```
davehunt: Rather than these if:assert blocks we should do something like `assert jobstatus in ('testfailed', 'busted', 'exception')`. This is much easier to read and would fail if the job status was anything else. Currently if the job status was another value, the test would pass.

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,23 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+
+
+@pytest.mark.nondestructive
+def test_view_unclassified_jobs(base_url, selenium):
+    page = TreeherderPage(selenium, base_url).open()
+    all_jobs = list(page.all_jobs)
+
+    page.filter_unclassified_jobs()
+    filtered_jobs = list(page.all_jobs)
+    jobstatus = page.result_sets[0].jobs[0].title
```
davehunt: We're just checking the first job status. While checking all jobs would probably be overkill, we could perhaps pick a few at random.

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,23 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+
+    assert not all_jobs == filtered_jobs
```
davehunt: I think we should examine this to ensure it's doing what we're expecting. It might be that these lists would not be equivalent even if the jobs were not filtered. I say this because if the DOM is rewritten then the elements will be different but with identical content. Maybe try replacing the filter action with a page refresh to see if this assert fails. If it it doesn't then we may need to do a deeper comparison of the lists.

pages/treeherder.py
MichalTHEDUDE: Hey @davehunt! Am I understanding correctly that there is no need to `return self`?
If so should we remove `return self` from `get_next_(ten/twenty/fifty)_results` function? Please check above rows 66, 71, 76.

tests/test_unclassified_job.py
MichalTHEDUDE: I'm trying to implement very the same logic (`jobstatus == page.result_sets[0].jobs[0].title`) and every time it gives me `IndexError: list index out of range`. 
IMO if there is an `if` statement this test will always pass. If we remove `if`, as you Dave suggesting, there will be an error.

I don't know if this `page.results_sets[0].jobs[0].title` is a good idea. What do you think?

pages/treeherder.py
davehunt: Yes, there's no value in returning the page object. If that code is already there then you can feel free to ignore it for this patch and perhaps remove it in a separate one.

tests/test_unclassified_job.py
davehunt: Index out of range suggests there are no jobs in the first result set. Is that the case? I don't see how that differs from your current code, which also accesses `page.result_sets[0].jobs[0].title`. We don't want a test that could never fail. :smile:

rbillings:
I addressed all of the comments except two. First: I wasn't sure how to compare all_jobs and filtered_jobs would for equivalency. The default all_jobs total will change as new jobs are run- but the amount of time it takes for the test to run is fast enough that that shouldn't happen. Same with filtered_jobs.
Second: I didn't change the assert format because it makes the test fail. If I assert that jobstatus is in the status [busted,etc.] it will compare the full title of the job to the status and fail. If you do individual asserts that say that the status is in the job title it will pass.

rbillings:
@davehunt please add to your review queue.

tests/test_unclassified_job.py
```diff
@@ -26,3 +25,23 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+
+
+@pytest.mark.nondestructive
+def test_view_unclassified_jobs(base_url, selenium):
+    page = TreeherderPage(selenium, base_url).open()
+    all_jobs = page.all_jobs
+
+    page.filter_unclassified_jobs()
+    filtered_jobs = page.all_jobs
+    jobstatus = page.select_random_job()
```
davehunt: This doesn't return a status but a job 'title'. Perhaps the property should be 'status' instead? Also, it doesn't select a job, it just picks a job at random and returns the title/status. This isn't a problem, but it shouldn't be in the page object - it should be in the test itself. See my other comment.

tests/test_unclassified_job.py
davehunt: We still need to remove these `if` blocks. Effectively what we're doing here is `if True: assert True` or `if False: assert True`. In the `False` case the assertion is never made. This way, the test could never fail. Instead we should check that one of the acceptable values is in the status. We can do this as:

```python
job = random.choice(filtered_jobs)
unclassified = ['testfailed', 'busted', 'exception']
assert any([status for status in unclassified if status in job.status])

**Comment by rbillings** [comment]
@davehunt comments have been addressed. I'll rebase after a review. Thanks

**Comment by rbillings** [comment]
@mozilla/firefox-test-engineering - I would appreciate it if someone could review this.

File: tests/test_unclassified_job.py [hunk 16]
```diff
@@ -26,3 +26,17 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+def test_view_unclassified_jobs(base_url, selenium):
+    page = TreeherderPage(selenium, base_url).open()
+    all_jobs = page.all_jobs
+
+    page.filter_unclassified_jobs()
+    filtered_jobs = page.all_jobs
+
+    job = random.choice(filtered_jobs)
+    unclassified = ['testfailed', 'busted', 'exception']
+    assert any([status for status in unclassified if status in job.title])
```
davehunt: Apologies if this was based on a suggestion by me, but I think we can make this more readable:

```python
assert any(status in job.title for status in unclassified)
```

tests/test_unclassified_job.py
```diff
@@ -26,3 +26,17 @@ def test_open_unclassified_failure_log(base_url, selenium):
...
+    assert not all_jobs == filtered_jobs
```
davehunt: Nit: I'd move this to line 38 after we've performed the filter.

rbillings:
All comments addressed!

PR #84, opened 2016-09-08
Improved filter verification

rbillings:
Filter test with improved verification.

### DIFF
--- a/tests/test_filter_jobs.py
+++ b/tests/test_filter_jobs.py
@@ -11,16 +11,16 @@ from pages.treeherder import TreeherderPage
 def test_filter_jobs(base_url, selenium):
     """Open resultset page and filter for platform"""
     page = TreeherderPage(selenium, base_url).open()
-    all_builds = list(page.all_builds)
+    all_builds = page.all_builds
 
     platform = u'Linux'
     page.filter_by(platform)
-    filtered_builds = list(page.all_builds)
+    filtered_builds = page.all_builds
 
     buildname = page.result_sets[0].builds[0].platform_name
     assert platform in buildname
 
-    assert not all_builds == filtered_builds
+    assert all_builds > filtered_builds
 
     page.clear_filter()
     platform2 = u'Windows'
--- a/tests/test_filter_jobs.py
+++ b/tests/test_filter_jobs.py
@@ -17,7 +17,6 @@ def test_filter_jobs(base_url, selenium):
     page.filter_by(platform)
     filtered_builds = list(page.all_builds)
 
-
     buildname = page.result_sets[0].builds[0].platform_name
     assert platform in buildname
 
--- a/tests/test_filter_jobs.py
+++ b/tests/test_filter_jobs.py
@@ -11,10 +11,17 @@ from pages.treeherder import TreeherderPage
 def test_filter_jobs(base_url, selenium):
     """Open resultset page and filter for platform"""
     page = TreeherderPage(selenium, base_url).open()
-    platform = u'Linux'
+    all_builds = list(page.all_builds)
 
+    platform = u'Linux'
     page.filter_by(platform)
-    assert platform in page.result_sets[0].builds[0].platform_name
+    filtered_builds = list(page.all_builds)
+
+
+    buildname = page.result_sets[0].builds[0].platform_name
+    assert platform in buildname
+
+    assert not all_builds == filtered_builds
 
     page.clear_filter()
     platform2 = u'Windows'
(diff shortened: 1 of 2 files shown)


stephendonner:
Passes for me locally, and in a couple repeated ad-hoc jobs, so r+ from that perspective.  I assume there were earlier review comments (given the list of commits), but I can't see them, so can't tell whether they were addressed here.

Probably worth getting another r, but I think that's up to you?

tests/test_filter_jobs.py
```diff
@@ -11,10 +11,16 @@
 def test_filter_jobs(base_url, selenium):
     """Open resultset page and filter for platform"""
     page = TreeherderPage(selenium, base_url).open()
-    platform = u'Linux'
+    all_builds = list(page.all_builds)
```
davehunt: You shouldn't need to include `list()` as the page object already has this.

tests/test_filter_jobs.py
```diff
@@ -11,10 +11,16 @@
...
 
+    platform = u'Linux'
     page.filter_by(platform)
-    assert platform in page.result_sets[0].builds[0].platform_name
+    filtered_builds = list(page.all_builds)
```
davehunt: As above, you shouldn't need `list()`

tests/test_filter_jobs.py
```diff
@@ -11,10 +11,16 @@
...
+
+    buildname = page.result_sets[0].builds[0].platform_name
+    assert platform in buildname
```
davehunt: I personally think it's more readable as `assert platform in page.result_sets[0].builds[0].platform_name` because otherwise we're using 'build' and 'platform' interchangeably.

tests/test_filter_jobs.py
```diff
@@ -11,10 +11,16 @@
...
+
+    assert not all_builds == filtered_builds
```
davehunt: There's a lot that could go wrong here and this would still pass. Perhaps we could also compare the lengths of these lists (filtered should be less), or maybe even check that filtered builds is a subset of all builds.

rbillings:
Ready for review, @mozilla/web-qa-sorcerers!

tests/test_filter_jobs.py
```diff
@@ -11,10 +11,16 @@
...
-    platform = u'Linux'
+    all_builds = page.all_builds
 
+    platform = u'Linux'
     page.filter_by(platform)
-    assert platform in page.result_sets[0].builds[0].platform_name
+    filtered_builds = page.all_builds
+
+    buildname = page.result_sets[0].builds[0].platform_name
+    assert platform in buildname
```
davehunt: Let's just do `assert platform in page.result_sets[0].builds[0].platform_name` and avoid the `buildname` variable.

tests/test_filter_jobs.py
```diff
@@ -11,10 +11,16 @@
...
+
+    assert all_builds > filtered_builds
```
davehunt: This needs to check the length of the build lists, so it should be `assert len(all_builds) > len(filtered_builds)`. When trying this locally I get failures either because the builds are still loading (so all_builds is less than filtered_builds) or because both lists are the same length. This is because we're returning all builds regardless of whether they are displayed.

Unfortunately there's no handy 'filter-shown' class like there is for jobs, but one potential solution to this is to filter the list in the `builds` property of `ResultSet`, such as:

```python
class ResultSet(Region):
 
    _platform_locator = (By.CLASS_NAME, 'platform')
 
    @property
    def builds(self):
        return [self.Build(self.page, root=el) for el in self.find_elements(*self._platform_locator) if el.is_displayed]
```

rbillings: @davehunt, I tried this method and it failed as the len of all_builds equalled filtered_builds. Is there really any value of doing a count when the filter method does not change the number of result sets? To test that the filter works is much more dependent on the value of the build platform name rather than the count.

I've tried to think of different ways of counting the page values before and after a filter- and there's no guarantee that any value will be greater or equal to the original.

davehunt: Hmm.. this worked for me. Why would filtered builds equal all builds? That suggests a bug considering we're testing that the builds are being filtered. Or are you saying that the filter might not impact the builds that happen to be displayed on staging? If so, I really think we need to stop writing tests that are so heavily dependent on the data that happens to be available - they're either going to be flaky or too forgiving to provide value.

rbillings:
This test is dependent on data - it seems like a good candidate to close, and then revisit in the future.

PR #88, opened 2016-10-12
Fix tox `passenv` for local test running

MichalTHEDUDE:
adding 'DISPLAY' into `passenv`


stephendonner:
Thx, @MichalTHEDUDE - runs fine for me on OS X: https://pastebin.mozilla.org/8918172 (I compared with the latest run at https://webqa-ci.mozilla.com/view/Treeherder/job/treeherder.stage/225/)

r+

davehunt:
So this is only needed if you're running your tests in a virtual display, which is often preferred when it's available. It doesn't affect Jenkins because these tests run in Sauce Labs. If we were running the tests in Travis CI then we would also want this. Glad it resolved your issues @MichalTHEDUDE and I'm happy to include this is any of our Tox environments that require a GUI.

PR #89, opened 2016-10-13
Feature/cleaning

MichalTHEDUDE:
I did a small code refactoring including:
- Cleaning - alphabetizing locators, methods and properties.
- Adding @property into "checkbox_is_selected" filter panel methods; applying changes in the test.
- Removing `return self` from methods.

About removing `return self`. @davehunt I remember our conversation in #83 and as far as I understood it correctly, we should remove all `return self`. Or maybe just in some cases?

Thanks!


pages/perfherder.py
```diff
@@ -13,7 +13,6 @@ class PerfherderPage(Base):
 
     def wait_for_page_to_load(self):
         self.wait.until(lambda s: self.is_graph_chooser_displayed)
-        return self
```
davehunt: This one makes sense to keep as we're extending the `Page.wait_for_page_to_load` method from PyPOM, which returns the page object. See http://pypom.readthedocs.io/en/latest/_modules/pypom/page.html#Page.wait_for_page_to_load

pages/treeherder.py
```diff
@@ -18,18 +18,18 @@ class TreeherderPage(Base):
...
     _mozilla_central_repo_locator = (By.CSS_SELECTOR, '#th-global-navbar-top a[href*="mozilla-central"]')
     _next_ten_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(1)')
     _next_twenty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(2)')
     _next_fifty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(3)')
-    _quick_filter_locator = (By.ID, 'quick-filter')
     _notification_locator = (By.CSS_SELECTOR, 'ul#notification-box')
+    _quick_filter_locator = (By.ID, 'quick-filter')
     _repos_menu_locator = (By.ID, 'repoLabel')
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
     _unchecked_repos_links_locator = (By.CSS_SELECTOR, '#repoLabel + .dropdown-menu .dropdown-checkbox:not([checked]) + .dropdown-link')
```
davehunt: Some projects group locators by logical blocks, but as we're not doing that here I'm +1 for sorting alphabetically!

pages/treeherder.py
```diff
@@ -38,7 +38,6 @@ class TreeherderPage(Base):
 
     def wait_for_page_to_load(self):
         self.wait.until(lambda s: self.unclassified_failure_count > 0)
-        return self
```
davehunt: As above, let's keep this one.

pages/treeherder.py
```diff
@@ -49,6 +48,18 @@ def all_jobs(self):
         return list(itertools.chain.from_iterable([r.jobs for r in self.result_sets]))
 
     @property
+    def checkbox_busted_is_selected(self):
+        return self.find_element(*self._filter_panel_busted_failures_locator).is_selected()
```
davehunt: +1 for making these properties!

pages/treeherder.py
```diff
@@ -99,6 +117,18 @@ def filter_by(self, term):
...
 
+    def get_next_ten_results(self):
+        self.find_element(*self._next_ten_locator).click()
+        self.wait.until(lambda s: len(self.result_sets) == 20)
+
+    def get_next_twenty_results(self):
+        self.find_element(*self._next_twenty_locator).click()
+        self.wait.until(lambda s: len(self.result_sets) == 30)
+
+    def get_next_fifty_results(self):
```
davehunt: So these aren't alphabetical, but in this case I prefer how you've sorted them by value.

pages/treeherder.py
```diff
@@ -291,27 +291,27 @@ def pin_job(self):
...
-        _root_locator = (By.ID, 'pinboard-panel')
-        _add_bug_locator = (By.CSS_SELECTOR, '#pinboard-related-bugs .pinboard-preload-txt')
         _add_bug_id_locator = (By.CSS_SELECTOR, '#related-bug-input')
+        _add_bug_locator = (By.CSS_SELECTOR, '#pinboard-related-bugs .pinboard-preload-txt')
         _clear_all_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-menu li:nth-child(4)')
-        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .save-btn-dropdown')
         _jobs_locator = (By.CLASS_NAME, 'pinned-job')
+        _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .save-btn-dropdown')
         _pinboard_remove_job_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job-close-btn')
+        _root_locator = (By.ID, 'pinboard-panel')
```
davehunt: I prefer to have `_root_locator` at the top as it's a special case.

pages/treeherder.py
```diff
@@ -350,7 +350,6 @@ class LogviewerPage(Page):
 
     def wait_for_page_to_load(self):
         self.wait.until(lambda s: self.is_job_status_visible)
-        return self
```
davehunt: Let's restore this one too.

pages/perfherder.py
MichalTHEDUDE: Thanks Dave! Because I was a little bit confused about that.
@rbillings here is the answer for my yesterdays question :)

pages/treeherder.py
MichalTHEDUDE: I know. But just wanted to be consistent with the numbers.

MichalTHEDUDE:
I've made a changes as you suggested.

pages/treeherder.py
```diff
@@ -120,66 +151,35 @@ def pin_using_spacebar(self):
...
-    def checkbox_testfailed_is_selected(self):
-        return self.find_element(*self._filter_panel_testfailed_failures_locator).is_selected()
-
-    def checkbox_busted_is_selected(self):
-        return self.find_element(*self._filter_panel_busted_failures_locator).is_selected()
+        self.find_element(*self._filter_panel_testfailed_failures_locator).click()
 
-    def checkbox_exception_is_selected(self):
-        return self.find_element(*self._filter_panel_exception_failures_locator).is_selected()
+    def select_random_repo(self):
```
MichalTHEDUDE: @rbillings is this method used somewhere?

rbillings: It is not currently used, however it was intended to be in test_switch_repo.py. There was a problem with the data dependency, so we hard coded in which repo to select - but also kept this method for future use.

Issue #52, opened 2016-04-27
Test: Filter by job keywords


rbillings:
Create a test for filtering jobs by keywords.

1) Open treeherder
2) Select any job
3) Within the job details panel on the left hand side, click any of the keyword links next to "Job"

Verify that list of jobs updates to display only jobs with the selected keyword[s]. Verify that filter displays in the header.

MichalTHEDUDE:
Hi Guys! I can work on this one!

- By 'keyword links next to "Job"' you mean like: 'linux opt', 'linux x64 pgo' or 'Windows XP opt' right?
- By 'Verify that filter displays in the header' you mean in the page URL after clicking?

rbillings:
Next to "Job" in the attached screenshot is a link with keywords. A better verification would be that the filtered jobs match the keywords from the link.

ex of link: https://treeherder.allizom.org/#/jobs?repo=mozilla-inbound&selectedJob=34980321&filter-searchStr=Linux%20x64%20asan%20Reftests%20executed%20by%20TaskCluster%20desktop-test-linux64-asan%2Fopt-reftest-no-accel-4%20tc-R(Ru4)

<img width="346" alt="screen shot 2016-10-07 at 10 09 24 am" src="https://cloud.githubusercontent.com/assets/737878/19198882/2e2fe0cc-8c76-11e6-8984-cb211225b113.png">

rbillings:
This was done.

PR #90, opened 2016-10-15
Test: Filter by job keywords

MichalTHEDUDE:
Test for Issue #52.
Waiting for the feedback and review :)


rbillings:
The code looks correct. My only concern is that by only checking the build platform, this test really isn't any different from test_filter_jobs. There are more keywords on the job details panel that are outside of the build platform.

MichalTHEDUDE:
Hmm, okay so I will try to check keywords on a job details panel.

MichalTHEDUDE:
@rbillings could you please check this?

I was thinking about adding two `select_random_job()` and `assert` but I don't know it it make sense.

PR #91, opened 2016-10-19
Results visible

rbillings:
Added new locator for asserting the result sets have loaded in the page. I also updated the log in locator, and added it for add_new_jobs as that task now requires log in.


pages/treeherder.py
```diff
@@ -29,6 +29,7 @@ class TreeherderPage(Base):
     _next_twenty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(2)')
     _next_fifty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(3)')
     _notification_locator = (By.CSS_SELECTOR, 'ul#notification-box')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden.ready')
```
rbillings: What is not in alphabetical order?

rbillings: You are correct- I didn't see it because of the line the comment was on. I will fix it [I like alphabetical also!]

rbillings:
Ready for review! @davehunt , @MichalTHEDUDE , @stephendonner , @m8ttyB

MichalTHEDUDE:
Question about this `results_visible` property. 
What is the rule of applying this assertion? Isn't that this way, that we can insert it to any test?

rbillings:
@MichalTHEDUDE This property has been added to the treeherder code, so yes you can insert it into any test.

pages/treeherder.py
```diff
@@ -86,6 +87,10 @@ def result_sets(self):
         return [self.ResultSet(self, el) for el in self.find_elements(*self._result_sets_locator)]
 
     @property
+    def results_visible(self):
+        return self.find_element(*self._resultset_visible_locator)
```
davehunt: Would it make sense to put this into `wait_for_page_to_load`? That way we'll be implicitly checking this by waiting for it to be true every time we instantiate a page. Also, we can add an explicit call anywhere a page load is expected. For example, put `self.wait_for_page_to_load()` into `get_next_ten_results` instead of adding asserts to the tests.

rbillings:
@davehunt I did move the results_visible locator, that's a good idea! I removed the asserts in most tests, however I did not do it for get_next_results as the results_visible locator appears when any number of results are visible and those tests rely on waiting for a particular number to be visible.

pages/treeherder.py
```diff
@@ -28,16 +28,18 @@ class TreeherderPage(Base):
     _next_ten_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(1)')
     _next_twenty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(2)')
     _next_fifty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(3)')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden.ready')
```
MichalTHEDUDE: Do we need 2 `_resultset_visible_locator`s? :)

pages/treeherder.py
```diff
@@ -28,16 +28,18 @@ class TreeherderPage(Base):
...
     _unchecked_repos_links_locator = (By.CSS_SELECTOR, '#repoLabel + .dropdown-menu .dropdown-checkbox:not([checked]) + .dropdown-link')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
     _unclassified_failure_filter_locator = (By.CSS_SELECTOR, '.btn-unclassified-failures')
 
     def wait_for_page_to_load(self):
-        self.wait.until(lambda s: self.unclassified_failure_count > 0)
-        return self
+        el = self.find_element(*self._resultset_visible_locator)
+        self.wait.until(EC.visibility_of(el))
```
MichalTHEDUDE: I've used this implementation and I've got `TimeoutException`, while I was trying to use `page.wait_for_page_to_load()`.

Did you have the same?

tests/test_unclassified_job.py
```diff
@@ -11,6 +11,7 @@
 def test_unclassified_failure(base_url, selenium):
     """Open resultset page and search for next unclassified failure"""
     page = TreeherderPage(selenium, base_url).open()
+    assert page.results_visible
```
davehunt: I believe this can now be removed?

tests/test_unclassified_job.py
```diff
@@ -22,6 +23,7 @@ def test_unclassified_failure(base_url, selenium):
 def test_open_unclassified_failure_log(base_url, selenium):
     """Open the job log and verify there is content"""
     treeherder_page = TreeherderPage(selenium, base_url).open()
+    assert treeherder_page.results_visible
```
davehunt: This can be removed now too?

rbillings:
Updated per comments. Ready for review.

pages/treeherder.py
```diff
@@ -28,16 +28,18 @@ class TreeherderPage(Base):
     _next_ten_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(1)')
     _next_twenty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(2)')
     _next_fifty_locator = (By.CSS_SELECTOR, 'div.btn:nth-child(3)')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden.ready')
     _quick_filter_locator = (By.ID, 'quick-filter')
     _repos_menu_locator = (By.ID, 'repoLabel')
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden.ready')
```
davehunt: This still appears twice as per @MichalTHEDUDE's comment.

pages/treeherder.py
davehunt: Are you still seeing issues with this @MichalTHEDUDE?

rbillings:
Fixed the duplicate locator- I blame rebasing! :P @MichalTHEDUDE I'll wait to hear from you. I have not seen any timeout exceptions from using the locator.

MichalTHEDUDE:
Just double checked that and still got TimeoutException.
I'm running `test_range.py` test when we can find `page.wait_for_page_to_load()`

Do you have the same error? Or something with my dev?

rbillings:
I've never seen that error. @davehunt - do you have any input on the error? Every other comment has been addressed.

davehunt:
> I've never seen that error. @davehunt - do you have any input on the error? Every other comment has been addressed.

I can replicate the same timeout as @MichalTHEDUDE - do you mean to wait for this element to be displayed? I can see it in the dom, but it's not displayed for me.

MichalTHEDUDE:
But should we use `span.hidden.ready`? Or it can be changed? 
Because if I try with `#global-container` it works like a charm.

Btw. Should there be `return self` at the end? (sorry still don't get it 100%)

davehunt:
> But should we use `span.hidden.ready`? Or it can be changed?

There could be a good reason for it - for example the application could be adding the `ready` class only when everything else is done. At least, I've seen this done before specifically for testing reasons.

> Btw. Should there be `return self` at the end?

You're right, there should. This allows us to call `wait_for_page_to_load` directly when instantiating a page object, such as:

```python
page = Page(selenium, base_url).wait_for_page_to_load()
```

rbillings:
The .ready is added to the hidden span to deliberately indicate when the resultset has loaded [see https://bugzilla.mozilla.org/show_bug.cgi?id=1263236]. My changes should hopefully make that more clear.

stephendonner:
I gave this several runs both locally and on ad-hoc jobs on staging Jenkins, and couldn't conclusively see that this added to stability, sorry.

Complicating that, though, is that there' s also at least one bug pre-existing on Treeherder staging, which is that the CPP and gl[n] links have errors - they are missing logfiles :-(

screen shot 2017-01-06 at 11 05 13 pm
screen shot 2017-01-06 at 11 04 54 pm

rbillings:
The log viewer issues are not related. The fix will land with this patch: https://github.com/taskcluster/taskcluster-worker/pull/149/commits/850159311ae85413269743f9a461d7d24340142b

stephendonner:
Heya @rbillings this is now out-of-sync with master, due to https://github.com/mozilla/treeherder-tests/commit/2d3a43724b1df50827fe9ca13f8d73d802b6dff1 being merged.  Mind updating when you get a chance?

I'd like to take another, closer look at this pull, as I'm learning Selenium now :-)

rbillings:
@stephendonner updated & ready when you are.

stephendonner:
I ran this both locally (Firefox 47/Mac) + on staging Jenkins' ad-hoc job, and found it really hard to tell vs. the baseline build from master (which, itself, varied from as high as 10 failures in a run to 5).

Sorry - thought I'd be a good candidate to review it, but I'm still ramping up quite a bit, so I'll defer to @davehunt and/or someone else, for help reviewing this properly!

(I'll still watch it and try to learn from others, of course.)

rbillings:
@mozilla/firefox-test-engineering I would appreciate a review!

pages/treeherder.py
```diff
@@ -31,12 +31,14 @@ class TreeherderPage(Base):
...
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden')
     _unchecked_repos_links_locator = (By.CSS_SELECTOR, '#repoLabel + .dropdown-menu .dropdown-checkbox:not([checked]) + .dropdown-link')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
     _unclassified_failure_filter_locator = (By.CSS_SELECTOR, '.btn-unclassified-failures')
 
     def wait_for_page_to_load(self):
-        self.wait.until(lambda s: self.unclassified_failure_count > 0)
+        el = self.find_element(*self._resultset_visible_locator)
+        self.wait.until(lambda s: 'hidden' in el.get_attribute('class'))
```
davehunt: The element is located based on the presence of the `hidden` class, so this wait will be immediately true. I'm also concerned that `span.hidden` is very vague and could easily locate other elements that may be added in the future.

Looking at the DOM, I wonder if a better locator might be `#th-global-content span.hidden.ready` (I suspect the `.ready` is important. Alternatively, you could us `#th-global-content span.hidden` and wait for the `class` to include `ready` or the `ng-class` to include `jobsReady`.

rbillings: I tried every variation I could think of for this with span.hidden and the .ready state, in addition waiting for the class to include ready. I also tried with ng-class. All of the tests continued to close the browser before jobs had loaded. Instead of waiting for the unclassified failure count in the header, I changed it to check that a result set has loaded in the page which more closely matches what #th-global-content span.hidden.ready is supposed to verify.

pages/treeherder.py
```diff
@@ -31,13 +31,13 @@ class TreeherderPage(Base):
...
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden')
     _unchecked_repos_links_locator = (By.CSS_SELECTOR, '#repoLabel + .dropdown-menu .dropdown-checkbox:not([checked]) + .dropdown-link')
     _unclassified_failure_count_locator = (By.ID, 'unclassified-failure-count')
     _unclassified_failure_filter_locator = (By.CSS_SELECTOR, '.btn-unclassified-failures')
 
     def wait_for_page_to_load(self):
-        self.wait.until(lambda s: self.unclassified_failure_count > 0)
-        return self
+        self.wait.until(lambda s: len(self.result_sets) >= 1)
```
davehunt: We still need to include `return self`

pages/treeherder.py
```diff
@@ -31,13 +31,13 @@ class TreeherderPage(Base):
     _quick_filter_locator = (By.ID, 'quick-filter')
     _repos_menu_locator = (By.ID, 'repoLabel')
     _result_sets_locator = (By.CSS_SELECTOR, '.result-set:not(.row)')
+    _resultset_visible_locator = (By.CSS_SELECTOR, 'span.hidden')
```
davehunt: This locator is now unused.

pages/treeherder.py
davehunt: This may return when one result set exists, even if more are yet to load. If that's okay, then r+ but I think it would be better if we could tell when the page has finished the initial load of results.

rbillings: That would be ideal, however I confirmed that the code for the span.hidden.ready only appears when a job button displays and not when the page loads. So there isn't a good way to verify the initial load that I know of.

rbillings:
Comments addressed, r?

Issue #54, opened 2016-04-27
Test: Filter jobs by email


rbillings:
Create two tests that add and remove filters by email address.

Filter by Email address
1) Open Treeherder page
2) Click a random email address from any of the jobs listed

Verify that jobs list is updated to only display jobs by selected author.

Remove email address filter
3) Click active repo name [default is mozilla-inbound]

Verify that jobs list is updated to display jobs by all authors.

MichalTHEDUDE:
I can work on this one :).

rbillings:
Great, thanks @MichalTHEDUDE

rbillings:
This was done.

PR #92, opened 2016-10-28
Test: Filter jobs by email

MichalTHEDUDE:
Creating two tests. 
First, filter by random email address and verify if jobs list show one author.
Second, remove the filter and verify that jobs list is updated to display jobs by all authors.

### DIFF
--- a/tests/test_filter_job_by_email.py
+++ b/tests/test_filter_job_by_email.py
@@ -28,7 +28,6 @@ def test_remove_email_address_filter(base_url, selenium):
     page.select_random_email()
 
     page.click_on_active_watched_repo()
-    first_email = page.random_email_name
-    second_email = page.random_email_name
+    all_emails = [email.get_name for email in page.all_emails]
 
-    assert first_email != second_email
+    assert len(set(all_emails)) > 1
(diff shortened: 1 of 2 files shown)


MichalTHEDUDE:
Please refer to Issue #54

tests/test_filter_job_by_email.py
```diff
@@ -0,0 +1,34 @@
...
+    """Open Treeherder page, select and filter by random email,
+    remove email address filter, verify that job list display all authors"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.select_random_email()
+
+    page.click_on_active_watched_repo()
+    first_email = page.random_email_name
+    second_email = page.random_email_name
+
+    assert first_email != second_email
```
rbillings: I just loaded Treeherder and there was one email address listed for multiple jobs- so while it's improbable, it's not impossible that 2 randomly selected emails will be for the same address. Perhaps it would be better to assert that the all_emails list contains more than just the first email found.

MichalTHEDUDE:
I've changed second function to check if length of this unique list is greater than 1.
I'm also thinking that using logic from function 2, function 1 can be refactored and we can delete `random_email_name` property.

What do you think?

PR #94, opened 2016-11-10
REMOVE all tests & code related to logging in

MichalTHEDUDE:
Remove all tests and code related to logging in into Persona.org (which will be shut down in a while).


stephendonner:
LGTM, too; r+ and merging!

Issue #57, opened 2016-05-04
Test: Keyboard shortcuts


rbillings:
There are quite a few keyboard shortcuts described here: https://treeherder.allizom.org/userguide.html

Create a test verifying that each shortcut works correctly. The "n" and spacebar keyboard shortcuts already exist and can be used as an example.

rbillings:
Next step: Check if there is coverage for this in the unit tests. If not, determine which keyboard shortcuts are not data dependent.

MichalTHEDUDE:
I can start working on this issue :)

rbillings:
This is completed for now- there is always room for future improvements.

PR #97, opened 2016-11-15
Feature/keyboard shortcuts

MichalTHEDUDE:
Testing the keyboard shortcuts as in Issue #57. 
It is just first part of a test and I will add more later.
----
Shortcuts that cannot be tested, due to logging in requirement are:
- Ctrl + Enter - save pinboard classification and related bugs
- r - retrigger selected job

----
Shortcuts already tested:
- n - select next unclassified failure
- l - open the logviewer for selected job
- spacebar - add a selected job to the pinboard


pages/treeherder.py
```diff
@@ -154,6 +159,11 @@ def display_keyboard_shortcuts(self):
         el = self.find_element(*self._result_sets_locator)
         el.send_keys(Keys.SHIFT + '?')
 
+    def filter_by_using_quick_filter(self, term):
```
MichalTHEDUDE: I don't know what to do with this.
Because it duplicate `filter_by` function.

pages/treeherder.py
```diff
@@ -122,6 +122,10 @@ def unclassified_failure_count(self):
     def clear_filter(self):
         self.selenium.find_element(*self._clear_filter_locator).click()
 
+    def clear_filter_by_shortcut(self):
```
MichalTHEDUDE: It duplicates `clear_filter` function, but using keyboard shortcut.
What to do with this??

MichalTHEDUDE:
3 tests left, which are not working at my site.
Tests:
- 'Ctrl + backspace' - delete job classification and related bugs
- 'a' - ignore unclassified failure lines
- 's' - save all autoclassifications

geckodriver.log
```diff
@@ -0,0 +1,371 @@
+1479297289077	geckodriver	INFO	Listening on 127.0.0.1:41593
```
davehunt: I'm sure you didn't mean to commit this. 😄

pages/treeherder.py
davehunt: I would maybe propose a `method` argument to the existing method, which defaults to 'pointer' but can also be 'keyboard'. What do you think?

pages/treeherder.py
```diff
@@ -91,15 +125,27 @@ def unclassified_failure_count(self):
...
+        el = self.find_element(*self._result_sets_locator)
+        el.send_keys(Keys.CONTROL + Keys.SHIFT + 'f')
```
davehunt: I would send keys to the `<body>` as that's more representative of what the user is likely to be doing.

pages/treeherder.py
```diff
@@ -91,15 +125,27 @@ def unclassified_failure_count(self):
...
+
     def click_on_filters_panel(self):
         self.find_element(*self._filter_panel_locator).click()
 
     def click_on_active_watched_repo(self):
         self.find_element(*self._active_watched_repo_locator).click()
 
+    def click_on_in_progress_button(self):
```
davehunt: This isn't clicking. Let's rename to indicate the action the user is taking.. does this show all in progress jobs? If so, name it `show_jobs_in_progress` for example. We could also use `method` here too if it's possible to do by pointer and keyboard.

pages/treeherder.py
```diff
@@ -116,6 +162,15 @@ def deselect_testfailed_failures(self):
         """Filters Panel must be opened"""
         self.find_element(*self._filter_panel_testfailed_failures_locator).click()
 
+    def display_keyboard_shortcuts(self):
+        el = self.find_element(*self._result_sets_locator)
+        el.send_keys(Keys.SHIFT + '?')
+
+    def filter_by_using_quick_filter(self, term):
```
davehunt: I think this can be combined with `filter_by` using the `method` approach.

pages/treeherder.py
```diff
@@ -152,6 +207,22 @@ def open_perfherder_page(self):
...
+
+    def pin_job_and_enter_classification(self, classification):
+        el = self.find_element(*self._result_sets_locator)
+        el.send_keys('c')
+        el = self.find_element(*self._classification_comment_locator)
+        el.send_keys(classification + Keys.RETURN)
+
+    def pin_random_job(self, driver):
+        random_job = random.choice(self.all_jobs).click()
+        ActionChains(driver).key_down(Keys.CONTROL).click(random_job).perform()
```
davehunt: I think this should be a `pin` method on the `Job` object. Then the random choice is made in the test instead of the page object.

pages/treeherder.py
```diff
@@ -152,6 +207,22 @@ def open_perfherder_page(self):
     def open_repos_menu(self):
         self.find_element(*self._repos_menu_locator).click()
 
+    def pin_job_and_enter_bug_number(self, bug_number):
```
davehunt: It's not clear to me from the name of the method what's happening here.. which job is this pinning?

pages/treeherder.py
```diff
@@ -309,13 +415,24 @@ def pin_job(self):
...
         _root_locator = (By.ID, 'pinboard-panel')
+        _classification_comment_locator = (By.ID, 'classification-comment')
         _clear_all_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .dropdown-menu li:nth-child(4)')
         _jobs_locator = (By.CLASS_NAME, 'pinned-job')
         _open_save_menu_locator = (By.CSS_SELECTOR, '#pinboard-controls .save-btn-dropdown')
         _pinboard_remove_job_locator = (By.CSS_SELECTOR, '#pinned-job-list .pinned-job-close-btn')
+        _pinboard_related_bugs_locator = (By.CSS_SELECTOR, '.pinboard-related-bugs-btn a em')
+
+        @property
+        def is_classification_comment_not_empty(self):
```
davehunt: I would avoid naming methods `is_x_not_y` and instead use `is_x_y`. Then just alter your assertion to be `assert not page.is_x_y`.

pages/treeherder.py
```diff
@@ -309,13 +415,24 @@ def pin_job(self):
...
+            el = self.find_element(*self._classification_comment_locator)
+            return 'ng-not-empty' in el.get_attribute('class')
```
davehunt: This implementation detail is concerning. If we changed the name of this class, the test will probably never fail. I would be more comfortable if we had at least one test for each potential return value from this property.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import random
+
+from pages.treeherder import TreeherderPage
+
+
+def test_close_open_panels(base_url, selenium):
+    """Shortcut: 'esc'
```
davehunt: First line of the docstring should be a summary. See https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings for details.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    """Shortcut: 'f'
+    Open Treeherder page, verify if search box is empty, enter search box
+    filter using 'f' shortcut, type 'mozilla', verify if filter box contain
+    word mozilla"""
+    page = TreeherderPage(selenium, base_url).open()
+    assert page.get_search_box_text == ''
+
+    page.filter_by_using_quick_filter('mozilla')
+
+    assert page.get_search_box_text == 'mozilla'
```
davehunt: I'd name this `page.search_term`.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    page.filter_by_using_quick_filter('mozilla')
+    assert page.get_search_box_text == 'mozilla'
+
+    page.clear_filter_by_shortcut()
+
+    assert page.get_search_box_text == ''
+
+
+def test_next_job_shortcut(base_url, selenium):
+    """Shortcut: 'Rigth Arrow'
```
davehunt: Nit: Typo

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    Open Treeherder page, select random job, select job next to it and take
+    job keyword, go back to previous job and select next job using Right arrow
+    shortcut, verify if job keyword match"""
+    page = TreeherderPage(selenium, base_url).open()
+    all_jobs = page.all_jobs
+
+    # Check number of jobs
+    num_of_jobs = len(all_jobs) - 1
+    rnd_number = random.randint(0, num_of_jobs)
```
davehunt: It's good practice to avoid abbreviating variable names. I would suggest `random_index` or similar.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    next_job = rnd_number + 1
+
+    if rnd_number == num_of_jobs:
+        next_job = 0
```
davehunt: Why are you setting it to zero? Is there a reason the next job can't be the last job?

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # Select random job and job next to it
+    all_jobs[rnd_number].click()
```
davehunt: I don't think you need this initial click.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    all_jobs[next_job].click()
+    page.job_details.wait_for_region_to_load()
+    assumed_job_keyword = page.job_details.job_keyword_name
+
+    all_jobs[rnd_number].click()
+    page.job_details.wait_for_region_to_load()
```
davehunt: Can we move all the waits into the methods that cause them to be needed?

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    page = TreeherderPage(selenium, base_url).open()
+    all_jobs = page.all_jobs
+
+    # Check number of jobs
+    num_of_jobs = len(all_jobs) - 1
+    rnd_number = random.randint(0, num_of_jobs)
+    previous_job = rnd_number - 1
+
+    if rnd_number == 0:
+        previous_job = num_of_jobs
```
davehunt: Again I'm not sure this is needed.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    Open Treeherder page, show only unclassified failures, select random
+    failure, select previous failure, take job keyword, go back to next failure
+    and select previous unclassified failure using 'p' button shortcut,
+    verify if job keywords match"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.show_only_unclassified_failures()
+    all_unclassified_failures = page.all_jobs
+
+    # Check number of unclassified failures
+    num_of_unclass_failures = len(all_unclassified_failures) - 1
```
davehunt: Nit: Avoid unnecessary abbreviations.

davehunt: You shouldn't remove one from this, because it makes the variable name unrepresentative of the value. Instead, remove one when you use the value on the next line.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    rnd_number = random.randint(0, num_of_unclass_failures)
```
davehunt: Make this between 1 and the number of unclassified failures minus 1 to avoid the previous failure being -1.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    previous_failure = rnd_number - 1
+
+    if rnd_number == 0:
+        previous_failure = num_of_unclass_failures
```
davehunt: If you wanted, you could have additional tests for the boundaries, to make sure they behave correctly. Otherwise this is a use case you'll only randomly test, making the tests even less deterministic.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    """Shortcut: 't'
+    Open Treeherder page, select random job, get active tab name, depending on
+    tab header name verify if next tab have assumed value, change tabs using
+    't' keyboard shortcut"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.select_random_job()
+
+    active_tab_name = page.job_details.active_tab_name
+
+    if active_tab_name == 'Job details':
```
davehunt: This test should be deterministic. We should expect the first tab to be job details. Otherwise all the if blocks will be skipped and the test would pass. You could make the tab names a list, and simply loop through them.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+
+def test_toggle_pinning_job_during_clicking_shortcut(base_url, selenium, driver):
+    """Shortcut: CTRL/CMD + job
+    Open Treeherder page, pin two random jobs, verify if number of pinned jobs
+    is greater that 1"""
+    page = TreeherderPage(selenium, base_url).open()
+    assert 0 == len(page.pinboard.jobs)
+
+    page.pin_random_job(driver)
+    page.pin_random_job(driver)
```
davehunt: This could pin the same job twice? Do we need to pin more than one?

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    page.select_random_job()
+
+    bug_number = u'1729'
+    page.pin_job_and_enter_bug_number(bug_number)
+
+    bug_number = u'87539319'
+    page.pin_job_and_enter_bug_number(bug_number)
+
+    assert len(page.pinboard.jobs) == 1
+    assert bug_number in page.pinboard.related_bugs
```
davehunt: If you make the bug numbers a list you can loop through them for the assignment and assertion, to check they're both added.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    enter classification text, verify if there is one pinned job and if
+    classification field have comment"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.select_random_job()
+
+    classification_text = u'mozilla'
+    page.pin_job_and_enter_classification(classification_text)
+
+    assert len(page.pinboard.jobs) == 1
+    assert page.pinboard.is_classification_comment_not_empty
```
davehunt: Why not assert that 'mozilla' is in the classification comment?

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+
+def test_clear_the_pinboard_shortcut(base_url, selenium, driver):
+    """Shortcut: CTRL + SHIFT + 'u'
+    Open Treeherder page, pin two random jobs, verify if number of pinned jobs
+    is greater that 1, clear the pinboard using CTRL + SHIFT + 'u' shortcut,
+    verify if number of pinned jobs is zero"""
+    page = TreeherderPage(selenium, base_url).open()
+
+    page.pin_random_job(driver)
+    page.pin_random_job(driver)
```
davehunt: This could pin the same job twice.

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+    Open Treeherder page, count number of in progress jobs (pending and
+    running), click on 'in progress' button using 'i' keyboard shortcut,
+    verify that number of in progress jobs is euqal to zero"""
+    page = TreeherderPage(selenium, base_url).open()
+
+    num_of_pending_jobs = len(page.all_pending_jobs)
+    num_of_running_jobs = len(page.all_running_jobs)
+
+    assert num_of_pending_jobs > 1
+    assert num_of_running_jobs > 1
```
davehunt: There's no need to assign variables, just assert directly:

```python
assert len(page.all_pending_jobs) > 0
assert len(page.all_running_jobs) > 0
```

tests/test_keyboard_shortcuts.py
```diff
@@ -0,0 +1,279 @@
...
+def test_show_only_unclassified_failures(base_url, selenium):
+    """Shortcut: 'u'
+    Open Treeherder page, show only unclassified failures using 'u' button,
+    verify number of unclassified failures"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.show_only_unclassified_failures()
+
+    unclassified_jobs = len(page.all_jobs)
+
+    assert unclassified_jobs == page.unclassified_failure_count
```
davehunt: Nice. 👍

pages/treeherder.py
rbillings: The pin_job_and_enter_bug_number test was removed as it required a log in. Although the shortcut opens the pinboard, I don't think this should be tested as the pinned job and bug cannot be saved.

tests/test_keyboard_shortcuts.py
rbillings: As I said above, although the shortcuts work the action of pinning a job with a bug number cannot be completed without signing in- so I would vote to remove this test.

tests/test_keyboard_shortcuts.py
rbillings: This is another item where the shortcut works but you cannot actually pin the job with a classification as the save method doesn't work. I vote to remove it.

tests/test_keyboard_shortcuts.py
rbillings: Nit: Greater than 1.

rbillings: Ah, that is correct- I see the definition is for a toggle using ctrl.

geckodriver.log
MichalTHEDUDE: lol, thats funny :D. I haven't noticed that ....

pages/treeherder.py
MichalTHEDUDE: Dave could you please, elaborate a little bit more about this `method`? Because I'm not getting this.

pages/treeherder.py
MichalTHEDUDE: I was also thinking about the name, that it is misleading. This shortcut can do both show/unshow - on/off.
And also about the `method` - could you please give me an example?

pages/treeherder.py
MichalTHEDUDE: We can pin and enter bug number, but it will only display on a website.
Should I remove the test and related code?

pages/treeherder.py
MichalTHEDUDE: I will move it to `JobDetails` section. And probably must apply those `method` thing that you mentioned earlier, because there is already `pin_job` method.

pages/treeherder.py
MichalTHEDUDE: what about `x_is_not_y`? :)
Then we do not have to change the name of below class.
What do you think?

tests/test_keyboard_shortcuts.py
MichalTHEDUDE: What I meant here is that: 
Assume that `random_job` is the last job. Then if we add `1`, there will be `IndexError: list index out of range` when we try to `click()` it.
That is why i'm setting this to`0` :)

Does it make sense?

pages/treeherder.py
davehunt: I would probably make an exception for body and not create a locator. The body is not something that's likely to change:

```python
self.find_element(By.CSS_SELECTOR, 'body').send_keys(Keys.CONTROL + Keys.SHIFT + 'f')
```

pages/treeherder.py
davehunt: Sure! I was thinking something like:

```python
 def clear_filter(self, method='pointer'):
    if method == 'pointer':
        self.selenium.find_element(*self._clear_filter_locator).click()
    elif method == 'keyboard':
        self.find_element(By.CSS_SELECTOR, 'body').send_keys(Keys.CONTROL + Keys.SHIFT + 'f')
    else:
        raise Exception('Unsupported method: {}'.format(method))
```

pages/treeherder.py
davehunt: You don't need to change the name of any class, just return the opposite, for example:

```python
return 'ng-not-empty' not in el.get_attribute('class')
```

tests/test_keyboard_shortcuts.py
davehunt: You're already deducting 1 from the number of jobs, so number of jobs + 1 would be the last job, but an out of range error. I think it's better to do this unless we specifically want to test boundaries.

tests/test_keyboard_shortcuts.py
MichalTHEDUDE: I'm sure. But IMO there may be a problem with boundaries. Because page would not load all the jobs (plus there are constantly changing) at the beginning. 
If we would count `len(page.all_jobs)` at the beginning of the test and at the end, the numbers will be different.

tests/test_keyboard_shortcuts.py
MichalTHEDUDE: But then in `if` statement the number of unclassified failures will be different.

tests/test_keyboard_shortcuts.py
MichalTHEDUDE: pin one job and `== 1`?

pages/treeherder.py
rbillings: You aren't actually pinning the job or the bug, you're just putting it on the board but not saving it - so yes, I would remove the test and related code.

MichalTHEDUDE: @davehunt What is your vote? Remove or leave?

tests/test_keyboard_shortcuts.py
MichalTHEDUDE: @davehunt What is your vote?

pages/treeherder.py
davehunt: I agree with @rbillings

MichalTHEDUDE: Just to clarify. I will remove all methods and code related to logging in.

pages/treeherder.py
MichalTHEDUDE: I'm not sure about moving this, because there is a problem with `driver` and with ActionChains.
Would it be better to leave it here? As `pin_using_spacebar`?

rbillings:
Hey @MichalTHEDUDE - will you have time to work on these?

rbillings:
@MichalTHEDUDE let us know if you are interested in finishing this up

rbillings:
I would like to get this PR closed out. Let us know if you'll be able to work on this anytime soon, @MichalTHEDUDE. I will close this PR in a week if I don't hear back or if you're no longer interested.

rbillings:
I'm closing this out. I plan to open a new PR with the code started here, so that the tests can get into production.

Issue #65, opened 2016-05-09
Test: Job status buttons


rbillings:
This test verifies the job status buttons in the nav bar.

1) Open Treeherder
2) Click the red button for: "failures" in the status bar
Verify red circle does not have dot in the middle, verify failed jobs do not display
3) Repeat for success, retry, usercancel, and in progress buttons
Verify circles do not have dot in the middle, jobs do not display
4) Click the light blue button for: "coalesced"
Verify button does have dot in the middle, jobs do display
5) Click the red button
Verify dot appears, jobs return to display

MichalTHEDUDE:
Hi,

I have checked job status according to your instructions.
Everything seems to be fine (as you described).

Does red button include 'exceptions'? Because I've noticed them displaying with failures.

rbillings:
@MichalTHEDUDE - this is an issue to create an automated test to follow the steps above. Yes, that does include exceptions, failures, and busted status. Glad to know it's still working as expected.

MichalTHEDUDE:
Quick question.
Am I understand correctly that this Issue require to write a CSS Selectors and functions in pages/treeherder.py and also creating new .py script in tests folder? Similar to #69 ?

rbillings:
Yes, this test will be written using Python/Selenium and the methods should be written in the treeherder.py. This test is taken, but if you're up for it there are other open tests waiting to be written.

PR #108, opened 2017-01-27
Job status buttons

rbillings:
Written for issue #65.


pages/treeherder.py
```diff
@@ -25,6 +25,12 @@ class TreeherderPage(Base):
     _filter_panel_reset_locator = (By.CSS_SELECTOR, '.pull-right span:nth-child(3)')
     _filter_panel_testfailed_failures_locator = (By.ID, 'testfailed')
     _mozilla_central_repo_locator = (By.CSS_SELECTOR, '#th-global-navbar-top a[href*="mozilla-central"]')
+    _nav_filter_coalesced_locator = (By.CSS_SELECTOR, '.btn-ltblue-filter-chicklet')
+    _nav_filter_failures_locator = (By.CSS_SELECTOR, '.btn-red-filter-chicklet')
+    _nav_filter_inprogress_locator = (By.CSS_SELECTOR, '.btn-dkgray-filter-chicklet')
+    _nav_filter_retry_locator = (By.CSS_SELECTOR, '.btn-dkblue-filter-chicklet')
+    _nav_filter_successes_locator = (By.CSS_SELECTOR, '.btn-green-filter-chicklet')
+    _nav_filter_usercancel_locator = (By.CSS_SELECTOR, '.btn-pink-filter-chicklet')
```
davehunt: We should locate these elements by the `title` rather than the `class`, as the class relates to the style of the element and not the functionality. If the color of any of these changed, it would cause a functional test to fail.

tests/test_status_results.py
```diff
@@ -0,0 +1,40 @@
...
+    """Open resultset page and verify job status buttons in the nav"""
+    page = TreeherderPage(selenium, base_url).open()
+    page.filter_job_failures()
+    page.filter_job_successes()
+    page.filter_job_retries()
+    page.filter_job_usercancel()
+    page.filter_job_in_progress()
+
+    all_jobs = page.all_jobs
+    unfiltered_jobs = len(page.all_jobs)
```
davehunt: This isn't unfiltered, because lines 15-19 are hiding all the jobs.

tests/test_status_results.py
```diff
@@ -0,0 +1,40 @@
...
+
+    page.filter_job_failures()
+    all_jobs = page.all_jobs
+    job = random.choice(all_jobs)
+    unclassified = ['testfailed', 'exception', 'busted']
```
davehunt: These are the result values for failures, so I wouldn't call this `unclassified`.

tests/test_status_results.py
```diff
@@ -0,0 +1,40 @@
...
+    assert any(status in job.title for status in unclassified)
+
+    page.filter_job_successes()
+    page.filter_job_retries()
+    page.filter_job_usercancel()
+    page.filter_job_in_progress()
+
+    all_jobs = page.all_jobs
+    filtered_jobs = len(page.all_jobs)
```
davehunt: This is now unfiltered, as we're seeing all the jobs.

tests/test_status_results.py
```diff
@@ -0,0 +1,40 @@
...
+    job = random.choice(all_jobs)
+    all = ['testfailed', 'exception', 'busted', 'success', 'retry', 'coalesced', 'running']
```
davehunt: `all` is reserved in Python

davehunt: We're not showing `coalesced` so I think including that is misleading.

pages/treeherder.py
rbillings: @davehunt I'm stuck trying to work out how to find the locator based on title without using the css_selector of the class- which uses the color name. Suggestions would be appreciated.

davehunt: The following work for me:

```python
_nav_filter_coalesced_locator = (By.CSS_SELECTOR, '.btn-nav-filter[title=coalesced]')
_nav_filter_failures_locator = (By.CSS_SELECTOR, '.btn-nav-filter[title=failures]')
```

tests/test_status_results.py
rbillings: The term "Unclassified" is from Treeherder itself - for example the unclassified failures total in the header includes jobs that have one of these as a status.

rbillings:
I'm not sure why it says commits are from 11 days ago unless it was from updating the branch. Comments were addressed, tests are split into filter categories.

rbillings:
Just as a clarification, there isn't a test for usercancel as the page displays result sets for that filter but no jobs.

tests/test_status_results.py
```diff
@@ -0,0 +1,97 @@
...
+    page = TreeherderPage(selenium, base_url).open()
+    page.filter_job_failures()
+    page.filter_job_retries()
+    page.filter_job_usercancel()
+    page.filter_job_in_progress()
+
+    all_jobs = page.all_jobs
+    job = random.choice(all_jobs)
+    jobstatus = 'success'
+    assert jobstatus in job.title
```
davehunt: As there's just one status we're looking for, you could check all jobs:

```python
assert all(map(lambda job: 'success' in job.title, page.all_jobs))
```

tests/test_status_results.py
```diff
@@ -0,0 +1,97 @@
...
+    page = TreeherderPage(selenium, base_url).open()
+    page.filter_job_failures()
+    page.filter_job_successes()
+    page.filter_job_usercancel()
+    page.filter_job_in_progress()
+
+    all_jobs = page.all_jobs
+    job = random.choice(all_jobs)
+    jobstatus = 'retry'
+    assert jobstatus in job.title
```
davehunt: As above, we can check all jobs easily here.

tests/test_status_results.py
```diff
@@ -0,0 +1,97 @@
...
+    page.filter_job_successes()
+    page.filter_job_retries()
+    page.filter_job_usercancel()
+    page.filter_job_coalesced()
+    page.filter_job_in_progress()
+
+    all_jobs = page.all_jobs
+    job = random.choice(all_jobs)
+    jobstatus = 'coalesced'
+    assert jobstatus in job.title
```
davehunt: As above, we can check all jobs easily here.

tests/test_status_results.py
```diff
@@ -0,0 +1,97 @@
...
+    page = TreeherderPage(selenium, base_url).open()
+    page.filter_job_failures()
+    page.filter_job_successes()
+    page.filter_job_retries()
+    page.filter_job_usercancel()
+
+    all_jobs = page.all_jobs
+    job = random.choice(all_jobs)
+    jobstatus = 'running'
+    assert jobstatus in job.title
```
davehunt: As above, we can check all jobs easily here.

rbillings: This was great as it unearthed something I didn't realize- that the in progress filter actually checks for a status of "running" OR "pending", so the test is better for it.

rbillings:
Comments addressed, r?

PR #119, opened 2017-03-01
Added shorcut tests for panels and quick filter

rbillings:
Issue #57

### DIFF
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -25,7 +25,6 @@ class TreeherderPage(Base):
     _filter_panel_locator = (By.CSS_SELECTOR, 'span.navbar-right > span:nth-child(4)')
     _filter_panel_reset_locator = (By.CSS_SELECTOR, '.pull-right span:nth-child(3)')
     _filter_panel_testfailed_failures_locator = (By.ID, 'testfailed')
-    _info_panel_content_locator = (By.ID, 'info-panel-content')
     _mozilla_central_repo_locator = (By.CSS_SELECTOR, '#th-global-navbar-top a[href*="mozilla-central"]')
     _nav_filter_coalesced_locator = (By.CSS_SELECTOR, '.btn-nav-filter[title=coalesced]')
     _nav_filter_failures_locator = (By.CSS_SELECTOR, '.btn-nav-filter[title=failures]')
@@ -77,13 +76,12 @@ class TreeherderPage(Base):
         return el.is_displayed()
 
     @property
-    def job_details(self):
-        return self.JobDetails(self)
+    def info_panel(self):
+        return self.InfoPanel(self)
 
     @property
-    def info_panel_is_open(self):
-        el = self.find_element(*self._info_panel_content_locator)
-        return el.is_displayed()
+    def job_details(self):
+        return self.JobDetails(self)
 
     @property
     def nav_filter_coalesced_is_selected(self):
@@ -379,6 +377,15 @@ class TreeherderPage(Base):
             def click(self):
                 self.root.click()
                 self.wait.until(lambda _: self.page.job_details.job_result_status)
+
+    class InfoPanel(Region):
+
+        _info_panel_content_locator = (By.ID, 'info-panel-content')
+
+        @property
+        def info_panel_is_open(self):
+            el = self.find_element(*self._info_panel_content_locator)
+            return el.is_displayed()
 
     class JobDetails(Region):
 
--- a/pages/treeherder.py
+++ b/pages/treeherder.py
@@ -19,11 +19,13 @@ class TreeherderPage(Base):
     _clear_filter_locator = (By.ID, 'quick-filter-clear-button')
     _close_the_job_panel_locator = (By.CSS_SELECTOR, '.info-panel-navbar-controls > li:nth-child(2)')
     _filter_panel_all_failures_locator = (By.CSS_SELECTOR, '.pull-right input')
+    _filter_panel_body_locator = (By.CSS_SELECTOR, '.th-top-nav-options-panel')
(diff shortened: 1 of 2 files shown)


tests/test_keyboard_shortcuts.py
```diff
@@ -7,6 +7,53 @@
...
+    Open Treeherder page, open Filters panel, select random job, close all
+    panels using 'esc' button, verify if all panels are closed.
+    """
+    page = TreeherderPage(selenium, base_url).open()
+
+    page.click_on_filters_panel()
+    page.select_random_job()
+
+    assert page.filter_panel_is_open
+    assert page.info_panel_is_open
```
davehunt: I would use region objects for the filter panel and info panel, to prevent the page object becoming overloaded. Also, there's already a job details region, which is within the info panel.

tests/test_keyboard_shortcuts.py
```diff
@@ -7,6 +7,53 @@
...
+
+    assert not page.filter_panel_is_open
+    assert not page.info_panel_is_open
+
+
+def test_enter_quick_filter_shortcut(base_url, selenium):
+    """Open Treeherder, verify shortcut: 'f' moves cursor to filter search box.
+    Open Treeherder page, verify if search box is empty, enter search box
+    filter using 'f' shortcut, type 'debug', verify if filter box contains
+    word Mozilla.
```
davehunt: Shouldn't we be verifying 'debug' and not 'Mozilla'?

tests/test_keyboard_shortcuts.py
```diff
@@ -7,6 +7,53 @@
...
+def test_clear_the_quick_filter_shortcut(base_url, selenium):
+    """Open Treeherder, verify shortcut: CTRL + SHIFT + 'f' clears the filter search box.
+    Open Treeherder page, filter by 'Mozdebugilla', verify if filter box contains the
+    word 'debug', clear the quick filter using CTRL + SHIFT + f shortcut,
+    verify if search box is empty.
+    """
+    page = TreeherderPage(selenium, base_url).open()
+
+    page.filter_by('debug', method='keyboard')
+    assert page.search_term == 'debug'
```
davehunt: This is a duplicate assert from another test, so we don't really need it here.

tests/test_keyboard_shortcuts.py
```diff
@@ -7,6 +7,53 @@
...
+
+def test_clear_the_quick_filter_shortcut(base_url, selenium):
+    """Open Treeherder, verify shortcut: CTRL + SHIFT + 'f' clears the filter search box.
+    Open Treeherder page, filter by 'Mozdebugilla', verify if filter box contains the
+    word 'debug', clear the quick filter using CTRL + SHIFT + f shortcut,
+    verify if search box is empty.
+    """
+    page = TreeherderPage(selenium, base_url).open()
+
+    page.filter_by('debug', method='keyboard')
```
davehunt: We don't need to use the keyboard to perform the filter, only the clear filter.

tests/test_keyboard_shortcuts.py
```diff
@@ -7,6 +7,53 @@
...
+    page = TreeherderPage(selenium, base_url).open()
+    assert page.search_term == ''
+
+    page.filter_by('debug', method='keyboard')
+    assert page.search_term == 'debug'
+
+
+def test_clear_the_quick_filter_shortcut(base_url, selenium):
+    """Open Treeherder, verify shortcut: CTRL + SHIFT + 'f' clears the filter search box.
+    Open Treeherder page, filter by 'Mozdebugilla', verify if filter box contains the
```
davehunt: typo: Mozdebugilla

tests/test_keyboard_shortcuts.py
rbillings: Agreed. I considered adding the job_details under the info_panel as they cover a similar area, but as it's not a repeating page object I decided not to- let me know if you disagree.

pages/treeherder.py
```diff
@@ -380,6 +378,15 @@ def click(self):
                 self.root.click()
                 self.wait.until(lambda _: self.page.job_details.job_result_status)
 
+    class InfoPanel(Region):
+
+        _info_panel_content_locator = (By.ID, 'info-panel-content')
```
davehunt: I'd suggest this as the `_root_locator` for the region.

pages/treeherder.py
```diff
@@ -380,6 +378,15 @@ def click(self):
...
+
+        @property
+        def info_panel_is_open(self):
```
davehunt: Rename to simply `is_open`

pages/treeherder.py
```diff
@@ -380,6 +378,15 @@ def click(self):
...
+            el = self.find_element(*self._info_panel_content_locator)
+            return el.is_displayed()
```
davehunt: Unless we need to use the element in multiple places, or we're considering line lengths, I think it's better to just return the value directly. Note that the following assumes making the existing locator the root of the element as mentioned above.

```python
return self.root.is_displayed()
```

tests/test_keyboard_shortcuts.py
davehunt: I think I would put job_details within info_panel, but let's not increase the scope of this patch. Maybe add a TODO comment in JobDetails to move it into the InfoPanel region, as it would require some tests to be modified.

rbillings:
Ready for review, @mozilla/firefox-test-engineering !

rbillings:
For whatever reason I didn't get the suggested updates in the github view- I will complete those and update this PR.

rbillings:
OK, now I've completed the suggested edits.

PR #122, opened 2017-03-17
Declarative + activeData

stephendonner:
@davehunt / @rbillings /cc @mozilla/firefox-test-engineering r?

Mostly-passing (same as prod) ad-hoc run here: http://webqa-ci-staging1.qa.scl3.mozilla.com:8080/job/treeherder.adhoc/114/console


Jenkinsfile
```diff
@@ -1,154 +1,64 @@
...
-            runTox(environment)
-          } catch(err) {
-            currentBuild.result = 'FAILURE'
-            throw err
-          } finally {
-            dir('results') {
-              stash environment
-            }
-          }
+    stage('Test') {
```
stephendonner: Guessing this is some funky line-ending or editor-spacing/tabs vs. spaces thing; will try to fix.

Jenkinsfile
```diff
@@ -1,154 +1,64 @@
...
+      environment {
+        /** See https://issues.jenkins-ci.org/browse/JENKINS-42857 - we'd like to expand this out into multi-line concatenations */
+        PYTEST_ADDOPTS = "--color=yes --driver=SauceLabs --variables=capabilities.json"
```
davehunt: Put this in the main `environment` section on line 19. We only need to have two sections when we're using a secret variables file.

Jenkinsfile
```diff
@@ -1,154 +1,64 @@
...
-            currentBuild.result = 'FAILURE'
-            throw err
-          } finally {
-            dir('results') {
-              stash environment
-            }
-          }
+    stage('Test') {
+      environment {
+        /** See https://issues.jenkins-ci.org/browse/JENKINS-42857 - we'd like to expand this out into multi-line concatenations */
```
davehunt: As mentioned with previous pull requests, let's fix the issue number here if we want to restore the multiline string in the future, though I think in a few days we may be able to revert this workaround.

Jenkinsfile
davehunt: Also please add `--tb=short`

PR #120, opened 2017-03-14
Move job_details region under info_panel

rbillings:
Tests pass except test_unclassified_job.py. This gives an error on switching the window that I don't know how to fix- the logviewer page does open correctly.

```
    def open_logviewer(self):
        self.find_element(*self._job_details_panel_locator).send_keys('l')
        window_handles = self.selenium.window_handles
        for handle in window_handles:
            self.selenium.switch_to.window(handle)
>       return LogviewerPage(self.selenium, self.page.base_url).wait_for_page_to_load()
E       AttributeError: 'InfoPanel' object has no attribute 'base_url'

pages/treeherder.py:414: AttributeError
------------------------------------------- pytest-selenium --------------------------------------------
URL: https://treeherder.allizom.org/logviewer.html#?job_id=78079209&repo=mozilla-inbound
================================= 1 failed, 2 passed in 17.98 seconds ==================================
ERROR: InvocationError: '/Users/rebeccabillings/github/treeherder-tests/.tox/tests/bin/pytest --junit-xml=results/tests.xml --html=results/tests.html --base-url https://treeherder.allizom.org tests/test_unclassified_job.py'
```

PR #123, opened 2017-03-21
View single result test fix

rbillings:
Depends on #120 , but this should fix the current Jenkins errors on test_view_single_result.py.


rbillings:
Ready for review @mozilla/firefox-test-engineering

PR #124, opened 2017-03-21
Fix broken Dropdown test

rbillings:
Depends on #120 , but this should fix test_treeherder_menu.py.

### DIFF
--- a/tests/test_treeherder_menu.py
+++ b/tests/test_treeherder_menu.py
@@ -12,13 +12,9 @@ def test_treeherder_dropdown(base_url, selenium):
     """Switch between Treeherder and Perfherder using header dropdown"""
     treeherder_page = TreeherderPage(selenium, base_url).open()
 
-    # Verify Treeherder loads
-    assert treeherder_page.result_sets > 0
-
     # Switch to Perfherder page
     perfherder_page = treeherder_page.open_perfherder_page()
     assert perfherder_page.is_graph_chooser_displayed
 
     # Return to Treeherder page
     treeherder_page = perfherder_page.open_treeherder_page()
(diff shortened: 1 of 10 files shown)


rbillings:
The unclassified failure count is now the last thing to load on the page, causing failures. This test updates it to show that at least one result set has loaded. Ready for review @mozilla/firefox-test-engineering .

tests/test_treeherder_menu.py
```diff
@@ -13,12 +13,12 @@ def test_treeherder_dropdown(base_url, selenium):
     treeherder_page = TreeherderPage(selenium, base_url).open()
 
     # Verify Treeherder loads
-    assert treeherder_page.unclassified_failure_count > 0
+    assert treeherder_page.result_sets > 0
```
davehunt: This is effectively identical to to the `wait_for_page_to_load` in `TreeherderPage`, which is implicitly called by the `open` method. I think we can safely remove this assertion and the comment.

tests/test_treeherder_menu.py
```diff
@@ -13,12 +13,12 @@ def test_treeherder_dropdown(base_url, selenium):
...
 
     # Switch to Perfherder page
     perfherder_page = treeherder_page.open_perfherder_page()
     assert perfherder_page.is_graph_chooser_displayed
 
     # Return to Treeherder page
     treeherder_page = perfherder_page.open_treeherder_page()
-    assert treeherder_page.unclassified_failure_count > 0
+    assert treeherder_page.result_sets > 0
```
davehunt: As above, this is just duplicating the wait that's already called so I would say it's safe to just remove this line.

PR #126, opened 2017-03-21
Fix flaky Unclassified jobs test

rbillings:
This test has become more flaky as the unclassified failure count is the last thing to update on the page load.

### DIFF
--- a/tests/test_unclassified_job.py
+++ b/tests/test_unclassified_job.py
@@ -4,7 +4,6 @@
 
 import pytest
 import random
-import time
 
 from pages.treeherder import TreeherderPage
 
--- a/tests/test_unclassified_job.py
+++ b/tests/test_unclassified_job.py
@@ -4,6 +4,7 @@
 
 import pytest
 import random
+import time
 
 from pages.treeherder import TreeherderPage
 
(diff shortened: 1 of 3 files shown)


rbillings:
Ready for review @mozilla/firefox-test-engineering

rbillings:
As mentioned in another PR, the unclassified failure count is now the last thing to load on the page- which is causing failures. This should fix it by making sure the count displays before closing the test.

tests/test_unclassified_job.py
```diff
@@ -12,7 +12,7 @@
 def test_unclassified_failure(base_url, selenium):
     """Open resultset page and search for next unclassified failure"""
     page = TreeherderPage(selenium, base_url).open()
-    assert page.unclassified_failure_count > 0
+    page.wait_for_unclassified_failures_to_load()
```
davehunt: I wonder if we even need this assert/wait. It seems these tests are more concerned with opening an unclassified failure, and that seems to work even before the count is finished loading.

rbillings:
In my local tests I watched this fail to find an unclassified job while the page was loading- but as the jenkins job just passed without this change I'm going to keep my eye on it for the next few days & may decide to just close the PR [or apply suggested changes].

rbillings:
As the Treeherder Jenkins job continues to pass, this PR has become irrelevant. Closing!

